PR #42, opened 2014-06-11
Added a new ``list_data_files`` function that can be used to search for files recursively

astrofrog:
In the package-template, we currently search for c files manually:

```
c_files = []
for root, dirs, files in os.walk(PACKAGENAME):
    for filename in files:
        if filename.endswith('.c'):
            c_files.append(
                os.path.join(
                    os.path.relpath(root, PACKAGENAME), filename))
```

we now also need to search for all files in ``packagename/data/`` recursively, so I thought it would be more convenient to write a function to do this.

I can add tests if we think this is a good idea.

cc @embray @mdboom

### DIFF
--- a/astropy_helpers/tests/test_utils.py
+++ b/astropy_helpers/tests/test_utils.py
@@ -18,7 +18,7 @@ def test_find_data_files(tmpdir):
 
     filenames = sorted(os.path.relpath(x, data.strpath) for x in filenames)
 
-    assert filenames[0] == 'data.dat'
-    assert filenames[1] == 'sub1/data.dat'
-    assert filenames[2] == 'sub1/sub3/data.dat'
-    assert filenames[3] == 'sub2/data.dat'
+    assert filenames[0] == os.path.join('data.dat')
+    assert filenames[1] == os.path.join('sub1', 'data.dat')
+    assert filenames[2] == os.path.join('sub1', 'sub3', 'data.dat')
+    assert filenames[3] == os.path.join('sub2', 'data.dat')
--- a/astropy_helpers/tests/test_utils.py
+++ b/astropy_helpers/tests/test_utils.py
@@ -1,0 +1,24 @@
+import os
+from ..utils import find_data_files
+
+
+def test_find_data_files(tmpdir):
+
+    data = tmpdir.mkdir('data')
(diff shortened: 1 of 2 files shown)


astrofrog:
@embray - I've now updated this to use the Python 3.5 recursive glob syntax (in fact, on Python 3.5 it will use the real glob). Does this seem like a good plan? If you are ok with this, I can add tests.

astropy_helpers/utils.py
```diff
@@ -660,3 +662,35 @@ def minversion(module, version, inclusive=True, version_path='__version__'):
...
+
+
+def find_data_files(package, pattern):
+    """
+    Include files matching ``pattern`` inside ``package``.
+
+    Parameters
+    ----------
+    package : str
+        The pacakge inside which to look for data files
```
embray: "pacakge"

embray:
Yeah, I think this fine.

astropy_helpers/utils.py
astrofrog: Fixed

astrofrog:
I added a unit test - this is now ready for review.

astropy_helpers/tests/test_utils.py
```diff
@@ -0,0 +1,24 @@
...
+            f.write('test')
+
+    filenames = find_data_files(data.strpath, '**/*.dat')
+
+    filenames = sorted(os.path.relpath(x, data.strpath) for x in filenames)
+
+    assert filenames[0] == 'data.dat'
+    assert filenames[1] == 'sub1/data.dat'
+    assert filenames[2] == 'sub1/sub3/data.dat'
+    assert filenames[3] == 'sub2/data.dat'
```
embray: should probably add something like `filenames = [f.replace(os.sep, '/') for f in filenames]`

astrofrog: Fixed - I just used ``os.path.join``

embray:
Needs to be rebased for some reason but then I'll merge.

astrofrog:
@embray - done. It was a conflict with ``classproperty`` which was added in the mean time.

embray:
Changelog entry?

embray:
I'll just add one manually, nbd.

PR #72, opened 2014-08-07
Added kselect/-k option to test runner

eteq:
This is the companion to astropy/astropy#2833 .  @embray, I'm not entirely sure the best way to handle this... Because the test runner is a *runtime* dependency of the test command, this will cause the test command to fail if the astropy version installed locally is too old.  How do you suggest we get around that?


eteq:
@embray -  if I understand correctly, now that astropy/astropy#4020 is in there's no need for this, right? It can all go in astropy/astropy#2833 ?

embray:
That sounds right.

PR #103, opened 2014-10-27
Emit a warning if a public object has no docstring

astrofrog:
This was originally https://github.com/astropy/astropy/pull/2136 - I don't think we should merge as-is but instead make it an opt-in option that can be set in ``conftest.py``. Then even if we never enable it by default in astropy core we can then enable it from time to time locally for testing.


astrofrog:
While this would have been nice, we now no longer use ``AstropyAutosummary`` for Sphinx 1.2 so this PR is moot. I'll open an issue to keep track of the feature request: https://github.com/astropy/astropy-helpers/issues/195

PR #128, opened 2015-01-22
Enable MacOS X testing

astrofrog:
Also, make sure we create a conda environment with the right Python version, otherwise we were testing all builds with Python 2.7. Yes, you read this right, for the last few months the tests have only been running with Python 2.7 :-/ Oops! (sorry!)

This is pending Travis enabling the multi-os support for this repository (should be soon)


embray:
Fortunately I run all the tests locally with tox, it's not like Python 3 has been totally untested.

.travis.yml
```diff
@@ -1,11 +1,16 @@
-language: python
+language: c
```
embray: ?_?

astrofrog: The multi-os feature doesn't work if the language is set to Python. BUT actually we don't use the built-in Python since we install it with conda hence why this will work.

astrofrog:
**Note:** this may pass but the multi-os is not enabled yet, so we should wait and restart the build once it is.

.travis.yml
embray: Maybe just add a note about that in a comment?  I know I'll forget why this is what it is by the next time I see it :)

astrofrog: Done :)

astrofrog:
Homebrew will likely be needed to install things like graphviz.

astrofrog:
@embray - just for information, the Travis failure on MacOS X in Python 3 seems to be real and I have also had the same issue locally in the past (but I thought it was a fluke because it wasn't seen on Travis with Linux and Python 3). But it does seem to be a real Mac + Python 3 issue.

astrofrog:
It seems to be a Python 3.4 issue as opposed to just Python 3. Python 3.3 is fine. Weird.

astrofrog:
I can reproduce locally, but it is a sporadic failure because the test passes one in five times. I tried adding a delay before the module is imported but no luck.

astrofrog:
Ignore my comment about Python 3.3, it fails sporadically on either Python 3 versions

astrofrog:
We should worry about this after the 1.0 release candidate, it is just an issue with the testing not with the functionality as far as I can tell.

embray:
@astrofrog Are you referring to the `test_no_cython_buildext` that failed in the last build?  Yeah, I can't get that to fail on Linux.  If I were to throw out a hypothesis, it might have primarily to do with the filesystem time resolution.  A call to `importlib.invalidate_caches` after each `run_setup` call might fix it.  I'll see if I can find an OSX machine to test that on, or you can try it yourself by putting `import importlib; importlib.invalidate_caches()` just before it tries `import _eva_.unit01`.

astrofrog:
@embray - bingo :) I also had a hunch it was something like this but didn't realize you could invalidate the cache.

embray:
Well, that's a thing about the new import system implementation as of 3.2 or 3.3 (I can never remember).  Previously there was no such "cache", but now it caches the directory listings of `sys.path` entries, and the cache is only invalidated if the modtimes for any of those directories change.  So if you have a new module appear on the path within a second of when the import system was last invoked you can run into trouble.  This is especially problematic on OSX.  I actually submitted [a patch](http://bugs.python.org/issue17330) for this a while back, but it was ultimately rejected since it just pushed the issue into even more obscure territory.  And while I'm still not happy with the situation I agree with the decision to reject the patch for now.

astropy_helpers/tests/test_setup_helpers.py
```diff
@@ -10,6 +10,19 @@
...
+    """
+    In Python 3, on MacOS X, the import cache has to be invalidated otherwise
+    new extensions built with ``run_setup`` do not always get picked up.
+    """
+    try:
+        return run_setup(*args, **kwargs)
+    finally:
+        if sys.version_info[0] >= 3:
+            import importlib
+            importlib.invalidate_caches()
```
embray: Oh, my thinking was just to name this `run_setup` so that it would be transparent (I've had reasons to wrap `run_setup` in the past too, but have since removed them since they weren't necessary anymore, but now there's a new reason again.  I'll just update it myself.

astrofrog: Shall I fix this now or are you doing it manually?

embray: I'm doing it.

embray:
Merged in a12ae72061

astrofrog:
I'm confused, it looks like this was merged multiple times into master and includes the ``run_setup_patched`` from here (not ``run_setup``). Maybe something went wrong.

embray:
You know what. I think I know what I did.  I think I did a manual merge, but then clicked "Merge" in the PR instead of "Close".  Somehow, it miraculously still worked and did something weird.  I'll just fix this up manually in master.

embray:
I forced master back to a12ae72 which was my manual merge, and dropped the merge commit made through GitHub. Sorry for the confusion.

Issue #114, opened 2014-12-15
``setup.py`` fails if in a git repo with no commits (low priority)


astrofrog:
The following demonstrates the issue:

```
$ git clone git@github.com:astropy/package-template.git
Cloning into 'package-template'...
cd packaremote: Counting objects: 545, done.
remote: Total 545 (delta 0), reused 0 (delta 0)
Receiving objects: 100% (545/545), 160.67 KiB | 0 bytes/s, done.
Resolving deltas: 100% (273/273), done.
Checking connectivity... done.
mac-robitaille2:tmp tom$ cd package-template
mac-robitaille2:package-template tom$ rm -rf .git
mac-robitaille2:package-template tom$ git init

mac-robitaille2:package-template tom$ python setup.py
git submodule command failed unexpectedly:
error: pathspec 'astropy_helpers' did not match any file(s) known to git.
Did you forget to 'git add'?

The requested path 'astropy_helpers' for importing astropy_helpers does not exist, or does not contain a copy of the astropy_helpers package.  Attempting download instead.
Downloading astropy_helpers; run setup.py with the --offline option to force offline installation.
Traceback (most recent call last):
  File "setup.py", line 18, in <module>
    from astropy_helpers.setup_helpers import (
  File "/Users/tom/tmp/package-template/astropy_helpers-0.4.3-py3.4.egg/astropy_helpers/__init__.py", line 2, in <module>
    from .version import version as __version__
  File "/Users/tom/tmp/package-template/astropy_helpers-0.4.3-py3.4.egg/astropy_helpers/version.py", line 151, in <module>
    version = update_git_devstr(_last_generated_version)
  File "/Users/tom/tmp/package-template/astropy_helpers-0.4.3-py3.4.egg/astropy_helpers/version.py", line 34, in update_git_devstr
    devstr = get_git_devstr(sha=True, show_warning=False, path=path)
  File "/Users/tom/tmp/package-template/astropy_helpers-0.4.3-py3.4.egg/astropy_helpers/version.py", line 144, in get_git_devstr
    return _decode_stdio(stdout)[:40]
  File "/Users/tom/tmp/package-template/astropy_helpers-0.4.3-py3.4.egg/astropy_helpers/version.py", line 17, in _decode_stdio
    text = stream.decode(stdio_encoding)
AttributeError: 'str' object has no attribute 'decode'
```

While this may seem like a quite obscure case, it can happen when people are setting up affiliated packages from scratch (I ran into it).

[...]

Issue #107, opened 2014-11-24
packagename.__version__ has wrong git info when run in git repository


kbarbary:
To reproduce, cd into any git repository and import a *different* astropy affiliated package. If the affiliated package is on a development version, `packagename.__version__` will be "MAJOR.MINOR.devNUM" where "NUM" is wrong: it is the number of commits in the current git repository rather than that of `packagename`.

embray:
Should be fixed by #137.

Issue #53, opened 2014-07-07
Build Cython C sources as prereq of sdist


embray:
Somehow I was convinced this was already done, to ensure that C files compiled from Cython sources are included in the sdist.  But apparently that isn't working.  Something to do.

bsipocz:
sorry for closing, I was mistaken that this issue was in the package-template repo.

astrofrog:
As we approach end of life for astropy-helpers, we will never fix this so I will go ahead and close.

Issue #121, opened 2015-01-13
Confusing message from ah_bootstrap.py when installing from source dist when in a subdirectory of a git repo


embray:
The Astropy source distribution currently bundles astropy_helpers.  When installing a fresh copy of Astropy from source this results in the messages:

```
Downloading/unpacking astropy
  Downloading astropy-0.4.2.tar.gz (7.2MB): 7.2MB downloaded
  Running setup.py (path:/internal/1/root/src/astropy/astropy/.tox/py33/build/astropy/setup.py) egg_info for package astropy
    git submodule command failed unexpectedly:
    error: pathspec '.tox/py33/build/astropy/astropy_helpers' did not match any file(s) known to git.
    Did you forget to 'git add'?
    
    
Requirement already satisfied (use --upgrade to upgrade): numpy>=1.5.1 in ./lib/python3.3/site-packages (from astropy)
Installing collected packages: astropy
  Running setup.py install for astropy
    git submodule command failed unexpectedly:
    error: pathspec '.tox/py33/build/astropy/astropy_helpers' did not match any file(s) known to git.
    Did you forget to 'git add'?
```

The error message appears twice--once when `./setup.py egg_info` is run, and once when `./setup.py install` is run.  This is only happening because I was installing Astropy in a tox environment under the `.tox` directory in a project, which happens to live under a git repo.  So this is somewhat related to #107 in that it has to do with astropy-helpers (in this case ah_bootstrap.py itself, I think) not correctly detecting whether it is in a git repo *for* the project being installed.

PR #137, opened 2015-02-10
Fixes to version.py

embray:
These are some mostly minor fixes to `version.py` generation.  But it will improve the generation of "static" version.pys that do not contain any git-related functions.  This also ensures that for release versions only the static version.py will be generated.

A later fix will add hooks to the `sdist` command to ensure that:
a) The package is always built before running sdist (see #53)
b) A static `version.py` will always be generated for creating an sdist.

But I want to make that fix for a later version I think.  It's too late to do that for 1.0 unless it really appears critical.  I don't think it will be though, with the setup.py fix (as long as the same fix is added to Astropy as well as to package-template.

Note also this PR branched off on #136.


embray:
Added to this an additional fix to address #121.

embray:
I need to double-check, but I believe this also fixes #107.

embray:
Confirmed, ce29656 also fixes #107.  I want to see if I can come up with a test for that though.

embray:
Furthermore, this also should fix #114.

eteq:
Also, this should fix astropy/astropy#3475, correct?

astrofrog:
@embray - is this good to go once AppVeyor passes?

embray:
So...close...

embray:
Rebased and added a changelog entry.  Should be no need to re-run the tests.

PR #139, opened 2015-02-13
Fixes to the auto-generated _compiler module

embray:
I noticed in testing for the Astropy release that `astropy.version.compiler` is empty.  That's because the generated version.py looks in `packagename._compiler`.  But there were problems with that file being created.  Also, astropy itself is using its own `astropy.utils._compiler` which I believe can be simply removed now that this feature comes from astropy-helpers.


embray:
If this passes I'm going to backport it to v0.4.x, release astropy-helpers v0.4.6, and *then* I'll release Astropy v0.4.5 :D

embray:
\o/

PR #141, opened 2015-02-17
githash = 'None' when running generate_version_py for the first time with uses_get=False

embray:
(I could have sworn I submitted this yesterday, but it seems to have vanished...)

The behavior when running `generate_version_py` with the `uses_git=False` flag isn't quite right.  This flag should *not* mean that git should not be used at all to generate the `version.py` module.  Rather, it just means that the generated `version.py` itself should be the "static" version that does not contain any code that runs git.

This means that in a fresh checkout of a release version of Astropy, running `setup.py` for the first time results in a `version.py` that contains `githash = 'None'` because it won't run git to determine the latest commit hash.  I got around this while making the Astropy v0.4.5 release by manually running `generate_version_py` once with `uses_git=True`, and then running it again with `uses_git=False`.  

Relatedly, I feel that running `generate_version_py` out of a release version, when a `version.py` already exists, should be a no-op.  It should not generate a newer `version.py` than the one that was bundled with the release.

### DIFF
--- a/astropy_helpers/tests/test_git_helpers.py
+++ b/astropy_helpers/tests/test_git_helpers.py
@@ -35,7 +35,7 @@ from astropy_helpers.version_helpers import generate_version_py
 if not RELEASE:
     VERSION += get_git_devstr(False)
 
-generate_version_py(NAME, VERSION, RELEASE, False)
+generate_version_py(NAME, VERSION, RELEASE, False, uses_git=not RELEASE)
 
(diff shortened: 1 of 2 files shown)


embray:
Attached a fix for this--would like to get this in before releasing 1.0 as it will be good to have for making the actual Astropy 1.0 release.

astrofrog:
:+1:

embray:
Oops--apparently I forgot to push the changelog entry when I made this PR, although it was in my original branch.

PR #142, opened 2015-02-22
Added instructions for OTF download from PyPI

astrofrog:
In glue, we want to make use of astropy-helpers only for the Sphinx extensions, so including it as a submodule is overkill. In that case, it's enough to simply include:

```python
from setuptools import Distribution
Distribution({'setup_requires': 'astropy-helpers'})
```

at the top of the ``conf.py`` script and it will get downloaded from ``PyPI`` and installed into a temporary local directory. This PR adds instructions related to this to the README.

@embray - maybe you can comment on whether there is a better way to do this?

The glue PR is here: https://github.com/glue-viz/glue/pull/566

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,9 @@
 astropy-helpers
 ===============
 
+About
+-----
+
 This project provides a Python package, ``astropy_helpers``, which includes
 many build, installation, and documentation-related tools used by the Astropy
 project, but packaged separately for use by other projects that wish to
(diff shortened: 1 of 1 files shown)


embray:
No, I don't want to suggest this.  They should still just use ah_bootstrap.py.  That doesn't require anything having to do with submodules.  Or it shouldn't anyways.  ah_bootstrap.py already does what this describes.

embray:
That, or if it's only for docs, they should just install astropy_helpers as a build dependency for their docs, like any other package containing sphinx extensions.

astrofrog:
@embray - in this case it's purely for docs. I'm all for the option of including it as a docs build dependency, but if people want to build the docs locally it means they will have an installed version of astropy-helpers and I thought we didn't like that? (or was that just a problem with old versions of setuptools)?

embray:
Neither.  I recommend against it for non-developers.  For the purposes of building docs there's nothing wrong with it.

embray:
On a broader note, there's been an effort (currently on hold) to develop an improved interface for managing build-time dependencies that doesn't rely on easy-install, so in the meantime I don't want to encourage widespread use of this sort of hack.

astrofrog:
Closing this since it is not recommended.

PR #144, opened 2015-02-25
Allow astropy-helpers to override itself

embray:
This should provide a workaround for astropy/astropy#3541

I haven't written a test for this yet, though it would be a relatively difficult one (not impossible though so I may try to add one later).  In the meantime I'm going to test this directly using testpypi.

In order for this change to fully fix astropy/astropy#3541 there will need to be two new releases of astropy-helpers: v0.4.8, and v1.0.1 each of which will include this fix.  When installing poppy, webbpsf, or any other similar affiliated package that has not yet updated to astropy-helpers 1.0 *and* that lists astropy in `setup_requires`, the ah_bootstrap will auto-upgrade astropy-helpers to v0.4.8.  This will include the necessary logic required so that when it in turn tries to install astropy, astropy's *own* ah_bootstrap can run, upgrade itself to astropy-helpers v1.0.1, and carry on with the installation.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,20 @@ astropy-helpers Changelog
 - Nothing changed yet.
 
 
+1.0.1 (unreleased)
+------------------
+
+- Improved the ``ah_bootstrap`` script's ability to override existing
+  installations of astropy-helpers with new versions in the context of
+  installing multiple packages simultaneously within the same Python
+  interpreter (e.g. when one package has in its ``setup_requires`` another
+  package that uses a different version of astropy-helpers. [#144]
+
+- Added a workaround to an issue in matplotlib that can, in rare cases, lead
+  to a crash when installing packages that import matplotlib at build time.
+  [#144]
+
+
 1.0 (2015-02-17)
 ----------------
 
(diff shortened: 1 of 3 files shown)


embray:
Although in the process of writing this I realized a flaw with the plan--the fix that allows astropy-helpers to replace a previous astropy-helpers is in ah_bootstrap.  But when astropy-helpers auto-upgrades it does not run the *new* ah_bootstrap from the upgraded version.  I *had* thought of this problem previously, but somehow convinced myself (incorrectly) that it wouldn't be an issue.  I'll have to see if I can find a way around that...

embray:
I left out an import statement previously causing all the tests to fail.  Fixed that and rebased.  Still haven't found a full solution to the problem of self-replacing with a version of astropy-helpers that does not already support self-replacing.

embray:
I'd like to write a test for this, but it's very hard.  Not impossible, but hard...

I cobbled together something on testpypi, however, that confirmed that the combination of an astropy-helpers 0.4.x update, plus a new Astropy release with an updated astropy-helpers 1.0.1, fixes the installation issue with webbpsf.

PR #146, opened 2015-03-02
Remove timestamp of the build to allow reproducible builds

olebole:
See [ReproducibleBuilds](https://wiki.debian.org/ReproducibleBuilds) on the Debian wiki, and the [the astroquery package](https://reproducible.debian.net/dbd/sid/amd64/astroquery_0.2.3+dfsg-1.debbindiff.html) as an example.


embray:
Okay, but in real world practice this is very useful for us, so I think removing this feature outright is too heavy a hammer.  A way to disable it might be preferable.

olebole:
Do you have a use case for this? I could see the only reason in debugging -- and then one could also just check the creation time of the `version.py` file.

embray:
It is useful for debugging.  In particular, useful for debugging other peoples' installations, when I have to try to figure out exactly what they did.

One thing I'm wondering though, is it updating the timestamp when installing from a tarball?  One possibility would be to just disable version.py generation entirely in the case, since the included one should suffice anyways.

embray:
@olebole Would an option that can be changed in setup.cfg, or somewhere like that, to disable the timestamp be helpful?

olebole:
Optimal would be a command line flag (like --reproducible). For a change in setup.cfg, I still need to do  a patch -- however, it would probably still be better than to patch `version_helpers.py` directly.

embray:
What if instead there were something like an optional "site.cfg" that can be added to modify/compose with setup.cfg?  Then it's just a matter of adding a site.cfg with the appropriate settings, which is still preferable to a patch.  I believe that's how numpy and scipy handle site-specific build issues like these.

olebole:
On Debian, I always start with an unmodified base system, then install all the required packages, unpack the sources and finally apply any patches to the sources. Therefore, I do not have the option to add a `site.cfg`; I would always do something locally. Therefore, I guess the best compromise would be a patch of `setup.cfg` (or the `--reproducible` option being even better).

embray:
@olebole thinking about this again--if the timestamp were not included (or at least, were fixed) for released versions, would this be a problem?

Or in other words, is the variable timestamp a problem for dev versions?  Do you need the reproducibility when testing all dev versions too, or just releases?

olebole:
That is just for released versions. Usually, I only upload these to Debian -- maybe prereleases (alpha, beta, RC) to experimental. But to Debian unstable/testing/stable will go only stable versions with a defined version number.

embray:
Got it--even for pre-releases they would include a pre-packaged version.py too with a fixed version.  So I think I can do away with the timestamps on these.  I've only found them useful for development sometimes.

eteq:
@embray - can you clarify if you implemented what you mentioned here?  I.e., should this be closed, or left open as a reminder to do that?

olebole:
I still apply this to the Debian package of astropy-helpers, and other packagers that care about reproducibility (Fedora) would probably also benefit.

astrofrog:
@olebole - would #341 work for you too? (it would be better in that it still leaves the date for non-packaging purposes)

olebole:
Agreed.

bsipocz:
#341 was just merged, so closing this now. Thank you @olebole, and sorry that this slipped though under the radar.

PR #151, opened 2015-03-27
Better HiDPI support.

mdboom:
Use SVG for logo and linkout, falling back to PNG if the browser doesn't
support it.


embray:
FWIW, see also #150 for update of the favicon with hi-res support.

PR #152, opened 2015-03-27
Fix CI builds

embray:
An ongoing PR to try to fix some of the issues with the CI builds...


embray:
There also seems to be a problem with appveyor.  I'll look at that in a subsequent commit.

embray:
The appveyor thing may have just been temporary, upon closer inspection.  Not sure though.

embray:
Well, still a failure on appveyor, but again seemingly random-ish.  It seems like we're getting regular timeouts with the graphviz install through Chocolatey.  I'm not sure what, if anything, to do about that.

PR #153, opened 2015-03-30
Improve get_compiler_version

embray:
This improves the still somewhat unreliable (but there's no real great way to do this anyways) `get_compiler_version` function to support more compilers that might have different flags to get their version info, and to return more of said version info (basically everything its version option returns to stdout (or in some cases stderr) by default.

It would be better if we could find out what the compiler is by first compiling the _compiler module and getting info out of that.  But that doesn't help here, since this function is used to figure out what the compiler is before we compile anything in the first place.  Maybe we can do away with that somehow though, since this was mostly just a workaround to an issue in older versions of OSX.

In the meantime, this allows the existing function to work with more compilers, and to return more info.  This change is needed for astropy/astropy#1489

### DIFF
--- a/astropy_helpers/setup_helpers.py
+++ b/astropy_helpers/setup_helpers.py
@@ -49,7 +49,8 @@ _module_state = {
     'registered_commands': None,
     'have_cython': False,
     'have_sphinx': False,
-    'package_cache': None
+    'package_cache': None,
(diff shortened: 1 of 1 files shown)


embray:
Okay, so coverage decreased a little, but not much to do testing-wise here.

Issue #74, opened 2014-08-08
Add generic support for submodules


eteq:
It came up in astropy/photutils#132 that it might be nice to be able to have git submodules available in an affiliated package.  However, the bar for use of submodules is very high: for good or for ill, most people I talk about it are afraid of them/don't understand how to use them.

Fortunately, `ah_bootstrap` has solved this problem already, as has the `ipython` `setup.py` : just update the submodules as part of the setup process.It would be great to put machinery like this in the `astropy-helpers`, so that affiliated packages and astropy can start using it. 

@embray, do you thing the `ah_bootstrap` code can be translated fairly easily into a `setup.py` command to do submodule updating  (a la ipython)? Or would it be better to lift some of it from ipython, or start from scratch?  (or I suppose "not do it at all" is another option)

keflavich:
I'd like to chime in again now that I understand what `ah_bootstrap.py` does: YES, I want this feature!  I have a submodule included in pyspeckit that doesn't clone or load by default, which leads to tests (which I have in a submodule) not loading.  

Astroquery will probably want this too, eventually - for now, it doesn't need any submodules, though.

astrofrog:
How about we make it a command-line option for ``setup.py`` then specific projects can turn it on by default in the ``setup.cfg`` if needed?

embray:
I don't want to add any more zany command-line options if it can be avoided, but as I mentioned earlier I think additional submodules could be listed in setup.cfg.

embray:
@keflavich Note, you can use `git clone --recursive` to clone a repo *and* initialize all its submodules.  I recommend just making an alias for that.

keflavich:
@embray - I can, but other people cloning my repositories won't by default.  And someone installing from source but not from a clone won't get access to the submodule at all.  Right?

embray:
>  And someone installing from source but not from a clone won't get access to the submodule at all

If by which you mean a source tarball or something, if you want the submodule included with it you'll need to put the right entries in the MANIFEST.in file.  I would instead lean on the side of *not* doing that and instead have the submodule (if it is a Python package, at least) listed as a dependency instead.  But there are some cases where that's not preferable (I, for example, would prefer not to actually bundle astropy-helpers with the source tarball, but enough people preferred having it that I had no choice but to bend on that).

cdeil:
By now I also have a use for this feature.

[...]

PR #154, opened 2015-04-02
Make `check_submodule` public.

mdboom:
Over in pyasdf, I found it useful to use the `check_submodule` function in `ah-bootstrap` to update the `asdf-standard` submodule.  (See spacetelescope/pyasdf#86).  Over there, I'm using an old version of ah_bootstrap.py, where it was pretty easy to do.  Now I see that it's become a member function and is harder to use independently.  This restores it as a free function and makes it public.  A follow-on (if this makes sense) will be to update the package template to describe how to use this to automatically update git submodules.

### DIFF
--- a/ah_bootstrap.py
+++ b/ah_bootstrap.py
@@ -151,6 +151,211 @@ CFG_OPTIONS = [
 ]
 
 
+def check_submodule(path, use_git=True, offline=False):
+    """
+    Check if the given path is a git submodule.
+
+    See the docstrings for ``_check_submodule_using_git`` and
+    ``_check_submodule_no_git`` for further details.
+    """
+
+    if (path is None or
+            (os.path.exists(path) and not os.path.isdir(path))):
+        return False
+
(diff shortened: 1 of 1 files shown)


embray:
:-1: in favor of a more general approach to auto-initializing additional submodules as discussed in #74

embray:
I should correct somewhat since there's nothing "non-general" about this.  I guess I'll mostly reiterate my preference for ah_bootstrap to not offer any kind of API of its own.  A function like this could also just as easily go into astropy_helpers.setup_helpers too.  ah_bootstrap is needed to bootstrap astropy-helpers, but it shouldn't be involved in bootstrapping other parts of the package...

mdboom:
Ok.  Just so I'm clear: would you be in favor of a PR that adds this to `astropy_helpers.setup_helpers`, then?

embray:
Sure...

mdboom:
Closed in favor of #155.

PR #155, opened 2015-04-03
Add `update_submodules` to `setup_helpers.py`

mdboom:
This adds the same functionality that `ah_bootstrap.py` uses to get `astropy_helpers`, and makes it available for other submodules.

This automatically finds all the submodules by searching through `.gitmodules`, and then, if necessary, inits and updates them.  It respects the `--no-git` and `--offline` flags, and warns when they may be preventing automatic things from happening.

I'm not crazy about how the `--no-git` and `--offline` flags are passed along to setup_helpers, but I couldn't think of a better way.

### DIFF
--- a/ah_bootstrap.py
+++ b/ah_bootstrap.py
@@ -50,10 +50,12 @@ except ImportError:
 if sys.version_info[0] < 3:
     _str_types = (str, unicode)
     _text_type = unicode
+    import __builtin__ as _builtins
     PY3 = False
 else:
     _str_types = (str, bytes)
     _text_type = str
+    import builtins as _builtins
     PY3 = True
 
 
@@ -209,6 +211,8 @@ class _Bootstrapper(object):
 
         auto_use = config.pop('auto_use', False)
         bootstrapper = cls(**config)
+
+        _builtins._astropy_helpers_config = config
 
         if auto_use:
             # Run the bootstrapper, otherwise the setup.py is using the old
--- a/astropy_helpers/setup_helpers.py
+++ b/astropy_helpers/setup_helpers.py
@@ -7,14 +7,19 @@ setup/build/packaging that are useful to astropy as a whole.
 from __future__ import absolute_import, print_function
 
 import collections
+import io
 import os
 import re
 import shlex
-import shutil
 import subprocess
 import sys
 import textwrap
 import traceback
+
+try:
+    from ConfigParser import RawConfigParser
+except ImportError:
+    from configparser import RawConfigParser
 
 from distutils import log, ccompiler, sysconfig
 from distutils.dist import Distribution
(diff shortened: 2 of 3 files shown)


ah_bootstrap.py
```diff
@@ -210,6 +212,8 @@ def main(cls, argv=None):
         auto_use = config.pop('auto_use', False)
         bootstrapper = cls(**config)
 
+        _builtins._astropy_helpers_config = config
```
embray: I didn't even think about trying to support these options.  But yeah, let's not do this.

I've been thinking of getting rid of those altogether.  They can already be edited in setup.cfg if need be.  Furthermore, I've been thinking of adding support for a "site.cfg" which, if present, could override settings in setup.cfg.  That way developers and downstream packagers who need to make tweaks can put them there instead.

mdboom: Ok.  So I guess what needs to be done is:

- [ ] add `setup.cfg` parsing support to `setup_helpers.py` and grab the values from there.
- [ ] deprecate support for the commandline options in ah_bootstrap (since they won't work with `setup_helpers.py`.
- [ ] update all the messages that refer to the commandline options to refer to the `setup.cfg` settings

Am I on the right track?

embray: Sounds good I think.  Let me have a quick look at the setup.cfg parsing.  I don't think we actually need to write anything for that that isn't already there. 

I think also there's currently an ah_bootstrap section, but I also want to add or change that to astropy-helpers, since there are some other config values related to it that it would be nice to have in a config file.

eteq:
:+1: to this addition - I find myself wanting this from time to time.

I can't remember exactly how easy (possible?) it is to do this with distutils, but is there any chance of using this command to get ``--no-git`` and ``--offline ``on the list of options that ``python setup.py --help`` would show? I still think it's rather problematic that the standard ``--help`` doesn't tell you about that.

embray:
I think it's a slippery slope--really if you want to manage git submodules one should use git commands for that.  

>  but is there any chance of using this command to get --no-git and --offlineon the list of options that python setup.py --help would show

It's not possible.  All this "bootstrap" stuff happens long before distutils or setuptools are even invoked, and is forgotten by then too.

embray:
I'll add--these options were mostly added just to make things easier for downstream packagers.  I almost never use them myself (with the occasional exception of `--no-git`).

jhunkeler:
I know this is an ancient issue, but I recently wanted to automatically initialize an external submodule in an affiliate package. However I was immediately discouraged when I realized astropy_helpers didn't support this behavior.

This would be _quite useful_ for people that need it. Not having this functionality potentially forces the developer to homebrew something (and whatever they come up with may not be very robust), or end up dropping submodules altogether in favor of implementing a different way of obtaining external code.

(If something similar to this has already been implemented upstream please tell me. :))

pllim:
Hmm. This needs rebasing but @mdboom is no longer involved in the development of this tool. So, someone else needs to take over and implement the checkbox items mentioned above...

embray:
I think since this PR was first submitted I refactored some things that should make it easier.   Mostly all that needs to be added is a config option for something like "addtional_submodules" which can take a list of submodule names to update/init.

A lot of the stuff in ah_bootstrap.py is specifically geared toward finding and using some installation of astropy_helpers depending on the context (for example, checking if we're in a git repository, in which it will try to load it from a git submodule).  

I don't know how much all that stuff applies to another project, if at all.  Mostly what I've seen is people just want to be able to update/init a git submodule automatically when they run `setup.py` for the first time.  That's a bit simpler.

astrofrog:
Since this went stale and would need to be started from scratch, I opened an issue here: https://github.com/astropy/astropy-helpers/issues/351 and am closing this.

Issue #73, opened 2014-08-08
Move test stuff to astropy-helpers if there's a need for it outside astropy-dependents


eteq:
In APE4 and the current master, the test runner and related framework are *not* in  `astropy-helpers`, but rather in `astropy`.  The justification from @embray in APE4 is that the tests have to be at runtime, so it makes sense to leave it in `astropy` given that affiliated packages will need it at runtime anyway.

But while working on #72 I realized an issue with that: the test runner and test command are now in separate packages, so they can't be kept in sync.  More generally, it's really confusing to figure out how they interplay when they are in different packages.  So maybe we should just move all the testing stuff to `astropy-helpers` too?

cc @astrofrog

eteq:
@embray @astrofrog - I wanted to report an oddity related to this issue that I just encountered.  Halotools (which uses the package template) was getting strange errors on travis while trying to run their test suite.  It turns out this is because they were using numpy v1.8 for the main part of the test suite, and the conda packages for astropy are numpy 1.9 *only* after v4.1.  So they were getting astropy v4.1, which is incompatible with the current astropy helpers (due to the addition of the ``repeat`` keyword for the test runner).  Having all of the test stuff in the helpers would get around such versioning issues.

embray:
Interesting...I'll have to take a closer look in a bit.  But I feel like when I added the `repeat` keyword I updated the test runner so that it wouldn't use unsupported arguments.  In other words, I thought I made a specific point of addressing this kind of issue.

embray:
@eteq Aha.  It appears that [the workaround I added for this](https://github.com/astropy/astropy-helpers/commit/1068e26) was only added to the v0.4.x branch on astropy-helpers, and is not in master.  I guess the thinking there was that I was going to do something "better" later.  But this solution is still better than nothing so I better copy it over...

eteq:
@embray - ok, that sounds good for that particular issue.

The main point of this issue is still relevant because it would have prevented this in the first place.  But I guess it's not particularly more urgent now than before...

embray:
It's just too bad I didn't copy the change over--then it would've just worked like I planned =D

embray:
Oops--this was addressed by #184 and related.  The test stuff is all back in astropy itself.  It could still be split out of astropy too if some package wanted to use it without depending on astropy.  But in that case it should be its own entity separate from both astropy and astropy-helpers.

PR #159, opened 2015-04-13
Fix adaptible test command

embray:
Fix for the issue I originally mentioned in https://github.com/astropy/astropy-helpers/issues/73#issuecomment-72558556; this is just a matter of porting the change originally made in 1068e261d292814156e6c21f4f1f0a97dcffacc3 to work in the v1.0 release series.

This is still sort of a temporary hack in place of a better solution, but at issue is the interface between the `setup.py test` command, which is provided by astropy-helpers, and the actual capabilities of the astropy test suite in different versions of astropy and/or the package-template being used for affiliated packages.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,12 @@ astropy-helpers Changelog
   prevented descriptor classes with a custom metaclass from being documented
   correctly. [#158]
 
+- Fixed interface between ``setup.py test`` command and different versions
+  of the Astropy test runner and/or affiliated package template.  This was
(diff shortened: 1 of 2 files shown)


embray:
Darn it, this was supposed to be included in v1.0.3, but I forgot to add it to the milestone and it got dropped *again*.  I'm just going to close it though--I'm working on a different solution to this that will be more sustainable in the long term.

embray:
Not sure why I haven't still closed this.

PR #160, opened 2015-04-22
Fix invalid HTML

mdboom:
Obviously all browsers don't really care or we would have noticed this before, but this offends my sense of compliance ;)


embray:
Not that I'm not all for this, but I thought HTML5 removed the requirement from XHTML for empty tags to have an explicit close slash.

embray:
Oh. Apparently the template *is* XHTML, not HTML5.  Carry on then...

PR #161, opened 2015-04-29
add build_docs command/alias

eteq:
Someone at the Lorentz center python meeting (I think @kbarbary or maybe @cmccully or perhaps @Cadair ?)  mentioned that they found the command ``build_sphinx`` to be confusing.  This basically just adds an alias  to ``build_shinx`` command called ``build_docs`` . 

I have some misgivings about this because it's two commands that do the same thing, but I think ``build_docs`` is much more natural for someone who might not understand what "sphinx" is, so I mildly think it's worth it.

cc @embray @astrofrog

### DIFF
--- a/astropy_helpers/commands/build_sphinx.py
+++ b/astropy_helpers/commands/build_sphinx.py
@@ -222,3 +222,7 @@ class AstropyBuildSphinx(SphinxBuildDoc):
             # prevent that from running.  But there's no other apparent way
             # to signal what the return code should be.
             sys.exit(retcode)
+
+
+class AstropyBuildDocs(AstropyBuildSphinx):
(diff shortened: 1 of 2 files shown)


Cadair:
I think this was me... It has always bugged me a little bit.

embray:
I'm okay with it.

embray:
No reason this can't go in a patch release.  Just add a changelog entry for v1.0.3 please.

embray:
A concurrent update in the Astropy docs is a good idea too.  I'd be in favor of build_sphinx -> build_docs all around.

astrofrog:
+1 from me too (but let's keep build_sphinx for a while to avoid breaking all the CI)

PR #164, opened 2015-05-08
Fix css for Sphinx 1.3

### LINKED ISSUE #163: Code links have turned grey
In the latest version of the Astropy and affiliated package docs, code and API links have now turned grey!

screen shot 2015-05-07 at 11 37 57 pm

screen shot 2015-05-07 at 11 38 06 pm

@mdboom: you adjusted the CSS recently, could it be due to that?

mdboom:
Fixes #163 so our CSS is compatible with Sphinx 1.3 (where all `<tt>` elements were changed to `<code>`).  This carefully tries not to break things for Sphinx 1.2 in the meantime.

### DIFF
--- a/astropy_helpers/sphinx/themes/bootstrap-astropy/static/bootstrap-astropy.css
+++ b/astropy_helpers/sphinx/themes/bootstrap-astropy/static/bootstrap-astropy.css
@@ -137,7 +137,6 @@ tt {
   font-family: monospace;
 }
 code {
-  color: rgba(0, 0, 0, 0.75);
   padding: 1px 3px;
 }
 pre {
@@ -160,7 +159,7 @@ img {
 }
 
 /* format inline code with a rounded box */
-tt {
+tt, code {
     margin: 0 2px;
     padding: 0 5px;
     border: 1px solid #ddd;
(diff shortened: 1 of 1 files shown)


mdboom:
@embray: This should probably be backported to whatever is being tracked for astropy 1.0.x maintenance branch, so that if we have any more 1.0.x releases or doc updates they won't be built incorrectly by readthedocs.

mdboom:
Appveyor seems like a false negative here.

embray:
Okay; I'll leave the failed appveyor build as-is, since there's nothing in this change that would be picked up by the tests anyways.

mdboom:
Does this need a corresponding update to astropy's submodule pointer?

astrofrog:
Actually I think we currently tell RTD to use the latest git version of astropy-helpers:

https://github.com/astropy/astropy/blob/master/docs/rtd-pip-requirements#L1

I can't remember why we need to explicitly tell RTD to install it, there was a reason why it couldn't use the submodule... Or maybe it was just to keep up with changes faster. I'll trigger a new RTD build.

astrofrog:
Fixed: http://docs.astropy.org/en/latest/

Thanks @mdboom!

PR #168, opened 2015-06-15
Don't specify protocol on font loading

mdboom:
This is a analogous to astropy/astropy.github.com#81


embray:
Why close?

Would this help with astropy/astropy.github.com#74?

mdboom:
Once I realized that `https://` and not `//` was the way to go, I closed this, since this is already using `https://`.

embray:
Ah, got it.

PR #169, opened 2015-07-02
Make the font of graphviz diagrams match the HTML

mdboom:
Before this change, the graphviz diagrams use Times New Roman, which doesn't match the font used in the rest of the docs.


embray:
Thanks!

PR #170, opened 2015-07-02
Use mathjax for math rendering on readthedocs

mdboom:
Using mathjax for rendering math is generally of higher quality (particularly on HiDPI displays) than pngmath (which uses LaTeX at build time to render PNG images).  The reason we don't use mathjax is because then the documentation can't be viewed locally without an internet connection (unless we include around 35MB of extra stuff with our documentation).

However, for readthedocs, we can get away with using MathJax, since viewing the docs on readthedocs doesn't work without an internet connection anyway.  (I don't think they have a feature to download docs for offline viewing, but if they did, obviously this would break that).

I haven't tested this -- I think the only way to test it is to merge it, see how readthedocs likes it for the "latest" build and revert it if there's a problem.  None of that will affect historical or stable builds of the docs, and it's unlikely to even break "latest" since it won't replace content unless the build succeeds.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,10 @@ astropy-helpers Changelog
 
 - The fonts in graphviz diagrams now match the font of the HTML content. [#169]
 
+- When the documentation is built on readthedocs.org, MathJax will be
+  used for math rendering.  When built elsewhere, the "pngmath"
+  extension is still used for math rendering. [#170]
+
 1.0.2 (2015-04-02)
 ------------------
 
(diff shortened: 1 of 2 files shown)


embray:
Sounds good. Though it needs to be rebased now (I think because I merged #169)

mdboom:
Rebased.

embray:
Awesome, thanks.

mdboom:
This is "working for me" [TM], for example, here:

https://astropy.readthedocs.org/en/latest/modeling/compound-models.html

Looks good with Chrome 43 and Firefox 38 on Linux.  Anyone able to check IE and Safari?  (I assume it's all fine because mathjax claims to be supported there, but just in case).

embray:
Yeah, looks fine to me.

embray:
A better acid test is: https://astropy.readthedocs.org/en/latest/units/index.html

PR #171, opened 2015-07-07
This decorator shouldn't always assume that the docstring exists at a…

embray:
…ll; even if we expect the docstring to exist it may not due to optimized compilation (see astropy/astropy#3923)


mdboom:
:+1:  Makes sense.

PR #173, opened 2015-07-16
accept cpp files as valid Cython-produced outputs

eteq:
This PR is prompted by a strange error @aphearin has been seeing in [halotools](https://github.com/astropy/halotools). The basic problem is that if you use `setup_package.py` to create a Cython extension that uses C++ (as opposed to just C), the extension builds the first time, but every subsequent time it fails with an ``error: [Errno 2] Could not find C file  ...``.

After some investigation I traces the problem to the astropy_helpers `build_ext` command, which checks for ".c" files if it thinks the extension has already been built.  But if the extension is in C++, this  fails incorrectly because the files are ".cpp", not ".c".  So this PR simply adds a check for ".cpp" files as well.

I *think* this is the only place .c is assumes.  There's one other possible place: the helpers command which auto-generates the cython extensions if it finds pyx with c next to it.  I think in this case we don't want to do that because normally you need to explicitly tell cython to use C++ rather than just C.  But it would be easy enough to add if others disagree.

cc @embray @mdboom, as people who've helped write some of the extension-building machinery.  Also, @embray, what's the easiest way to quickly get this to halotools once it's merged?  Would you prefer a quick release, or should we just point it straight to the merge commit on master?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,9 @@ astropy-helpers Changelog
 - Fix crash when importing astropy_helpers when running with ``python -OO``
   [#171]
 
+- The ``build`` and ``build_ext`` stages now correctly recognize the presence
+  of C++ files in Cython extensions (previously only vanilla C worked). [#173]
+
 
 1.0.2 (2015-04-02)
 ------------------
--- a/astropy_helpers/commands/build_ext.py
+++ b/astropy_helpers/commands/build_ext.py
@@ -125,9 +125,11 @@ def generate_build_ext_command(packagename, release):
                 if src.endswith('.pyx'):
                     pyxfn = src
                     cfn = src[:-4] + '.c'
+                    cppfn = src[:-4] + '.cpp'
                 elif src.endswith('.c'):
                     pyxfn = src[:-2] + '.pyx'
                     cfn = src
+                    cppfn = src
 
                 if not os.path.isfile(pyxfn):
                     continue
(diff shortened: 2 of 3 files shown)


mdboom:
>  There's one other possible place: the helpers command which auto-generates the cython extensions if it finds pyx with c next to it.

I think we should update that, too.  If the user writes a c++ cython extension without that change, doesn't that mean it doesn't get autogenerated?

Also, we probably want to update the package-template to include `*.cpp` in the `.gitignore` (I've never been a fan of putting `*.c` there, but that would at least be consistent):

https://github.com/astropy/package-template/blob/master/.gitignore#L10

embray:
I kind of like putting almost any files in gitignore that I don't intend to add manually.  But I can see how it would make it easy to forget to add a file too.  I agree with @mdboom that for now might as well add it .gitignore for consistency's sake.

I also want to refactor this code in general, but for the near term it's good to just get it working.  Good news is I was planning to make a release of astropy-helpers this week anyways, so if you can just update this PR, and also add a changelog entry, I can include it in the release.

eteq:
Alright, I just pushed up some commits to do as @mdboom suggested and to add a changelog entry as @embray suggested.  I also made astropy/package-template#132 to add to the gitignore.  Good to go?

embray:
:+1:

eteq:
Cool - I just tested this on halotools and it resolves the issue.  @aphearin - note that you'll want to update halotools to the new release once @embray does the release (1.0.3, I think it is?)

aphearin:
Roger and wilco. I'll keep an eye out for v1.0.3

embray:
@aphearin Out of curiosity, what code in halotools was affected by this?  Looking at the Cython files currently in its master branch, I don't see any that look like there would be any reason for them to compile to C++.  Is this in some other branch?

I wanted to make sure if I update this I'd have something to test the change in (since per [this comment](https://github.com/astropy/astropy-helpers/commit/c91b40f5a8fd21278011bf177f2cc93551c1bd4e#commitcomment-12233932) I think the merged fix still needs tweaking).

embray:
Oh, nevermind, I see where you were using `libcpp.vector`.

aphearin:
@embray - yes, that's right. 

On the Halotools end, a crude operational verification is this: when I run my test suite twice in succession, the second test run fails and exits because the compiled files are not detected, as described https://github.com/astropy/halotools/issues/97. 

@eteq commented above that when he updated the copy of astropy_helpers embedded in Halotools, that this bug goes away, but I have not tested this myself.

embray:
I just tested in on Friday myself, and confirmed that it works now.  It also works fine with #174 which is what I really wanted to check.

aphearin:
Yup, I implemented this in a Halotools branch I'm working on. Seems like everything is kosher.

PR #174, opened 2015-07-17
Slight tweak to ensure that the (possible) C++ filename associated wi…

embray:
…th a Cython source is always found.

Followup to #173.  I've tested this against halotools, but probably won't add any other specific tests for now (still making a PR to run the tests though).

### DIFF
--- a/astropy_helpers/commands/build_ext.py
+++ b/astropy_helpers/commands/build_ext.py
@@ -122,14 +122,11 @@ def generate_build_ext_command(packagename, release):
 
             # Replace .pyx with C-equivalents, unless c files are missing
             for jdx, src in enumerate(extension.sources):
-                if src.endswith('.pyx'):
-                    pyxfn = src
-                    cfn = src[:-4] + '.c'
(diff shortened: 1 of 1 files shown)


embray:
This has been a frequent problem with the Windows builds of late:
```
Chocolatey v0.9.9.8
Installing the following packages:
graphviz.portable
By installing you accept licenses for the packages.
 
graphviz.portable v2.38
 Attempt to get headers for http://www.graphviz.org/pub/graphviz/stable/windows/graphviz-2.38.zip failed.
   Exception calling "GetResponse" with "0" argument(s): "The operation has timed out"
 Downloading graphviz.portable 32 bit
   from 'http://www.graphviz.org/pub/graphviz/stable/windows/graphviz-2.38.zip'
 Exception calling "GetResponse" with "0" argument(s): "The operation has timed 
 out"
```

Pretty annoying.  I wonder if there's a way to instruct Chocolatey to use alternative mirrors for certain packages, since graphviz.org is apparently not reliable... :(

PR #175, opened 2015-07-21
Fix output of build_sphinx -w option in py 3.x

eteq:
In python 3.x, it turns out that using the ``-w`` option causes the output of ``build_sphinx`` to be a single-line string that looks like ``b'...'`` (see gammapy/gammapy#305 for an report about this issue).

The underlying problem is that the ``-w`` path in the ``build_sphinx`` command has to know if the build succeeded, so it saves the stdout using ``subprocess.PIPE``.  It then prints at the end.  This works fine in py 2.x, because the pipes yield strings, but in py 3.x, the pipes yield bytes.  So this PR fixes this by decoding the output as UTF-8.  I *think* that should be safe, but @embray or @mdboom might want to provide input as to whether there's any way to actually check if the output is truly UTF-8.

### DIFF
--- a/astropy_helpers/commands/build_sphinx.py
+++ b/astropy_helpers/commands/build_sphinx.py
@@ -181,7 +181,7 @@ class AstropyBuildSphinx(SphinxBuildDoc):
 
             stdolines = stdo.splitlines()
 
-            if b'build succeeded.' in stdolines:
+            if 'build succeeded.' in stdolines:
                 retcode = 0
             else:
                 retcode = 1
--- a/astropy_helpers/commands/build_sphinx.py
+++ b/astropy_helpers/commands/build_sphinx.py
@@ -174,7 +174,7 @@ class AstropyBuildSphinx(SphinxBuildDoc):
                                     stdin=subprocess.PIPE,
                                     stdout=subprocess.PIPE,
                                     stderr=subprocess.STDOUT)
-            stdo, stde = proc.communicate(subproccode.encode('utf-8'))
+            stdo, _ = proc.communicate(subproccode.encode('utf-8'))
             stdo = stdo.decode('utf-8')
 
             print(stdo)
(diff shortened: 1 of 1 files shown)


embray:
Yeah, should be fine.

Incidentally, this business of calling subprocess and decoding the stdout is repeated in lots of places in astropy_helpers.  I might grub around and see if there isn't already a utility function for that (and add one if there isn't...)

eteq:
@embray - yeah, that definitely makes sense.  Let me know if so, as I would probably use it in other places.

mapazarr:
@eteq Thanks for taking care of this. The fix works for me. Sorry for the confusion in the last post: you're right. I was not using the `-w` option. There is no difference in the output when using the `-w` option for python `2.7` with and without the patch. Or, when using the patch, between python `2.7` and `3.4`. So all fine from my side :-)

As for the question why I am using the `-w` option is in order to catch possible failures in the docs that survive the tests but would fail in the sphinx build during a PR (up to now I only have done PRs in `gammapy`).

eteq:
@mapazarr - gotcha.  FYI, the ``-w`` option only changes the *return code* that the ``build_sphinx`` command yields.  You still will see all the warnings even if you don't provide it.  But travis uses the return code to decide if a test "passed" or "failed" in a PR, so that's why it's necessary there.  If you're just looking at it yourself, you can always just read the message at the end that tells if there were any warnings.

So it sounds like this does indeed fix the originally reported problem - if the tests pass I think this is good to go.

cdeil:
Tests passed.

Probably a changelog entry should be added.

embray:
I don't think it's super important.  Just to make sure I understand correctly, all this is doing is making the output a little more readable, right?  I can't say I've ever even noticed this myself, or didn't care when I did (though I'm certainly all for fixing it).

cdeil:
@embray – Here's the difference in output on Python 2 and 3:
- Python 2: https://travis-ci.org/astroplanners/astroplan/jobs/74846421#L866
- Python 3: https://travis-ci.org/astroplanners/astroplan/jobs/74846422#L876

This PR fixes the Python 3 output.

@embray – Ready to merge?

embray:
Ah, I see.  Yes, this is much better thanks.

PR #178, opened 2015-08-16
Add cmdline option to override auto_use in setup.cfg

sargas:
This is a simple patch to add a command line option (`--disable-ah-bootstrap`) to override the `auto_use = True` setting in packages. This is useful for packaging up astropy_helpers-based software for distributions without needing to manually patch it (as in https://anonscm.debian.org/cgit/debian-astro/packages/python-astropy.git/tree/debian/patches/disable_helper.patch by @olebole)

### DIFF
--- a/ah_bootstrap.py
+++ b/ah_bootstrap.py
@@ -16,7 +16,8 @@ When this module is imported or otherwise executed it automatically calls a
 main function that attempts to read the project's setup.cfg file, which it
 checks for a configuration section called ``[ah_bootstrap]`` the presences of
 that section, and options therein, determine the next step taken:  If it
-contains an option called ``auto_use`` with a value of ``True``, it will
+contains an option called ``auto_use`` with a value of ``True`` and setup.py,
+was not run with the ``--disable-ah-bootstrap`` option, it will
 automatically call the main function of this module called
 `use_astropy_helpers` (see that function's docstring for full details).
 Otherwise no further action is taken (however,
@@ -275,6 +276,10 @@ class _Bootstrapper(object):
         if '--offline' in argv:
             config['offline'] = True
             argv.remove('--offline')
+
+        if '--disable-ah-bootstrap' in argv:
+            config['auto_use'] = False
+            argv.remove('auto_use')
 
         return config
 


olebole:
I'd support this, since it makes my work (and the work of people packaging the affiliated packages for Debian) easier, thus lowering the threshold to get them packaged. Thank you for the patch!
If it is decided that the patch does not make sense in the helpers package, I would apply it locally to the Debian python-astropy-helpers package.

sargas:
@olebole thanks, but this just changes the default ah_bootstrap.py file - every affliated package using `auto_use = True` (including astropy itself) would need to update their bundled copy of it to use this option. It doesn't actually change the installed astropy-helpers.

olebole:
You are right: my mistake :-)

embray:
My preference is for downstream packagers to just patch the setup.cfg to set `auto_use = False`.  I've also mentioned elsewhere that what I would prefer to see also is the option for a site-specific setup.cfg that packagers can use to override the defaults (this is what Numpy does for example).

olebole:
@embray I can live with any option -- the current situation has the advantage that I can use it to teach new packagers how to patch the source when packaging. Site specific setup.cfg (in /etc) would be also an alternative to a setup option.

embray:
> the current situation has the advantage that I can use it to teach new packagers how to patch the source when packaging

True, though having to patch is not entirely ideal ;)
When adding a site specific setup.cfg does that also require a patch (at least for the Debian workflow)?  I didn't mean anything that would go in /etc though--I thought it would just be something you put in the source tree that overrides settings in the base setup.cfg.

embray:
Though in retrospect, although I would still like the config file option too, I think this is a fairly straightforward approach and a very simple patch.  If you could just add a changelog entry about it under the v1.0.4 section I'll merge.

olebole:
@embray Everything that is changes in the source tree comes there via a patch. I also have no direct access to the home directory during the build. The idea would just be that I add a proper `/etc/astropy-helpers.cfg` (or similar) to the astropy-helpers package, so that under Debian one would always use the installed package. However, the use of patching `setup.cfg` in the source to teach newcomers was not meant ironically; it is anyway very often needed to patch something, and therefore a packager should know how to do this.

embray:
One thing I do find funny is that the `ah_bootstrap` script doesn't even *allow* you to use an already installed version of astropy-helpers.  On one level this was intentional--a user installing from PyPI with pip for example should generally not install astropy-helpers directly.  But in cases where it *does* happen to be installed in site-packages for whatever reason it should still be possible to find the copy that's there and use it when appropriate.

sargas:
@olebole If `ah_bootstrap` uses that file if it exists, then you've changed (potentially breaking) anyone using pip in a virtualenv to install a package that requires a newer astropy_helpers for some reason. A `--use-setupcfg /path/to/deb_build.cfg`-like option is still needed, although I don't see the advantage of a site file over switches though.

@embray what about a `--use-system-astropy-helpers` option to make the intent clear? I think most distros have a concept of build-only dependences (DEPEND in ebuild's, Build-Depends in deb's, BuildPreReq in rpm's) that are not needed once the package is built/installed. The use of ah_bootstrap as a workaround for a library that isn't used after building is only needed for pip.

embray:
I think a `--use-system-astropy-helpers` would be good (unfortunately lengthy, but okay for static build scripts and also congruent with similar options for bundled build dependencies.

astrofrog:
Since this PR has been stale for a couple of years, I'm going to close it in favor of an issue: https://github.com/astropy/astropy-helpers/issues/350

Issue #179, opened 2015-08-19
bootstrap-astropy.css messes up line numbering on sphinx code blocks


RiceMunk:
Currently working on the documentation for something I'm trying to get released as an astropy affiliated package, and I found a problem with how numbered code blocks are rendered using the astropy sphinx css stuff.

Here's a screenshot of what happens (at least with Chrome and Firefox) if I try to make a line-numbered code block in sphinx (using the code-block::-markup):
![sphinx_codeblock_wrong](https://cloud.githubusercontent.com/assets/401690/9363229/16c40582-469d-11e5-9116-a4457e23ea43.png)

The line numbering on that is not very useful, so it's a problem for my attempt at writing examples.

After a while of scrathing my head and learning how the element inspection thingy on Chrome works, I found out that this is because of something in bootstrap-astropy.css. Turns out that, for reasons I don't really understand, the line "white-space: pre-wrap;" in the pre-block (line 153 in bootstrap-astropy.css) is causing this problem. If I remove that line from the css, I get something a bit better-looking:
![sphinx_codeblock_lesswrong](https://cloud.githubusercontent.com/assets/401690/9363326/c3549d70-469d-11e5-8862-512ac2b74997.png)

Gonna modify the css in the thingy I'm working on to not cause the line numbering issue, but I figured you people would wanna know about this as well.

embray:
That's odd. I think none of the Astropy docs use line numbering, so it hadn't come up before.  I'm not sure why this happens either--thanks for the report.

RiceMunk:
No problem. I went ahead and made a pull request for the quick fix I made.

Not sure if that's the best way of going about getting rid of this issue but there you have it. See pull request #180.

embray:
Fixed by #180--thanks @RiceMunk

PR #180, opened 2015-08-19
Fix for issue #179

RiceMunk:
This should fix issue #179 in the way I described in the OP.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,8 @@ astropy-helpers Changelog
 1.0.4 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Fixed issue with the sphinx documentation css where the line numbers for code
+  blocks were not aligned with the code. [#179]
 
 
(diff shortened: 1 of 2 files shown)


embray:
Probably fine, but I want to see if I can figure out why that line was there in the first place.

embray:
Okay, looks like that's been there all along (and is also overriding the `white-space: pre` just before it anyways).  I'm going to try building the Astropy docs to see if this change makes anything obviously go awry, and if not we can use it (just add a changelog entry for v1.0.4).

RiceMunk:
Yeah, I was wondering if there were supposed to be two `white-space` entries there, but don't really know enough about CSS to know if that intentional.

Added the changelog entry now to the file I think it's supposed to be in.

embray:
Looks good--I did a test build of the docs and although it's impossible to check every page I don't see this having a whole lot of impact anywhere, so we'll go with it for now.

PR #182, opened 2015-09-02
Crash in adjust_compiler when usable compiler isn't found

embray:
This showed up via a Stack Overflow question: http://stackoverflow.com/questions/32354982/problems-installing-astropy

If the `get_compiler_version` function fails to find the right way to get the compiler version, it returns the string `'unknown'`.  But the regexp in `adjust_compiler` that checks the compiler version is a bytes regexp.  On Python 2 this is no problem, but on Python 3 it crashes if passed the string `'unknown'`.

This can be replicated by setting the CC environment variable to any program that just exits with a non-zero error code (regardless of arguments).

Note, even if not for this bug the installation would probably still fail if the user lacks a working compiler.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,9 @@ astropy-helpers Changelog
 - Fixed issue with the sphinx documentation css where the line numbers for code
   blocks were not aligned with the code. [#179]
 
+- Fixed a crash that could occur on Python 3 when a working C compiler isn't
+  found. [#182]
+
 
 1.0.3 (2015-07-22)
 ------------------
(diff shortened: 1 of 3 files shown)


embray:
Honestly the test for this code probably wasn't even worth it.  The `adjust_compiler` hack has so far only ever been needed to work around problem's with Apple's llvm-gcc, which isn't even common in the wild anymore.  Maybe the code can hang around for now, but it's becoming irrelevant I think.

That said, adding this test boosted the coverage of that file a good %20 or so.

mdboom:
That's some test!  I agree with everything you say -- this is increasingly irrelevant.  `llvm-gcc` was always an "interim hack" until `clang` caught up.  I can't remember which version of XCode this corresponded to...

PR #183, opened 2015-09-14
Squelch Numpy deprecation warnings when building Cython extensions

embray:
Cython currently makes use of some deprecated (as of Numpy 1.7) interfaces in Numpy's C API, which results in some annoying warnings when building.

We can't disable use of the deprecated API, yet, since Cython depends on it.  But we can squelch those warnings for now, which are not of particular use to us.


embray:
Had to rebase, but since the last commit contains "skip ci" the CI stuff isn't running :(

I've tested this change locally, however, and found no errors, unsurprisingly.

PR #185, opened 2015-09-15
Add Cython support to the build_ext command late in the setup process

embray:
This makes it possible to build extension modules using Cython even if Cython was not available at the beginning of the setup.py script.  This means that Cython can be added as a build-time dependency in setup_requires, and still be used later in the build process if it was installed locally via setup_requires.

I plan to use this for astropy/astropy#3229 which I also want to extend to add Cython to the build requirements (only for git checkouts though--never source releases).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,11 @@ astropy-helpers Changelog
   just ensures that any other imports that occur during build will have this
   flag set. [#191]
 
+- It is now possible to use Cython as a ``setup_requires`` build requirement,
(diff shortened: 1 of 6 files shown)


embray:
Going to set up a PR to astropy incorporating this change before merging it.

embray:
Actually, nevermind, that will be part of astropy/astropy#3229 so I just need to update there.

embray:
Rebased on top of #192, and also removed the "skip ci" from the last commit so that the CI will run.

PR #186, opened 2015-09-15
Followup to #183 to fix building with MSVC

embray:
Fixes builds broken by #183.


embray:
The above Travis-CI failure is unrelated.

embray:
Not sure why "coverage pending" from coveralls.  The report seems to exist over at their site, so something is going wrong with their push to GitHub I guess...?

PR #187, opened 2015-09-17
Test temp root

embray:
This is needed as part of the solution for astropy/astropy#4164, which can then be resolved by setting a default temp-root option in the setup.cfg for astropy (or other package).  Tempting to just make this the default, but leaving as is for now.

A separate PR will be needed to implement this for v1.1.0.


embray:
Oops, I sort of did this backwards.  Hold on...

embray:
I'm actually going to close this PR.  Per the discussion in astropy/astropy#4164 this may not be the best approach to fix the original issue.

I'm still going to keep #188--the new feature may be useful to have at some point.  But it can wait 'til the next feature release and is less imperative.

PR #188, opened 2015-09-17
Adds an option to set the base directory for temporary test files

embray:
(ports #187 to master)


mdboom:
py.test 2.8.0 just came out on 18-09-2015 with a lot more changes.  

https://pytest.org/latest/changelog.html#id1

We may want to move to it while we're at it and make sure there are no more surprises.  There are some failures on the astropy suite when moving to it, but I haven't had time to deeply investigate yet.

embray:
Perhaps, but I don't want it to hold things up, and I don't see anything in the changes that would help with the specific issue this was trying to address.

PR #189, opened 2015-09-21
Ensure common py.test rootdir for test files

embray:
This now copies the docs for testing into the temp dir, and ensures that the setup.cfg is in the common root directory for the Python files and the docs.  This ensures that in py.test >= 2.7 the correct common test directory can be found as the 'rootdir'.  Seems to work fine with older versions of py.test as well (down to 2.4).

This replaces #187 as a solution to astropy/astropy#4164


mdboom:
:+1:

mdboom:
Sorry for closing.  Bumped the wrong button :(

embray:
Happens :)

PR #192, opened 2015-09-29
Fixed a regression introduced by by #184 that caused testing to always default to verbose results

embray:
Fixed another small regression from #184.  This fix should be incorporated in astropy/astropy#4020


embray:
Coverage dropped, but only in a module that isn't covered by the tests in the first place (and should probably be excluded from the coverage report since it only exists for backward-compat purposes).

PR #193, opened 2015-09-30
Adds a coveragerc file

embray:
Add a coveragerc to tweak the coverage report somewhat.  Excludes a few files from the report (not making a huge impact on the overall coverage score, but the rest are things that should be tested...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ before_script:
     - git config --global user.email "author@example.com"
 
 script:
-    - py.test --cov astropy_helpers astropy_helpers
+    - py.test --cov astropy_helpers --cov-config astropy_helpers/tests/coveragerc astropy_helpers
 
 after_success:
     - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,8 @@ before_script:
     - git config --global user.email "author@example.com"
 
 script:
-    - py.test --cov astropy_helpers --cov-config astropy_helpers/tests/coveragerc astropy_helpers
+    # Use full path for coveragerc; see issue #193
+    - py.test --cov astropy_helpers --cov-config $(readlink -f astropy_helpers/tests/coveragerc) astropy_helpers
 
 after_success:
     - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ before_script:
 
 script:
     # Use full path for coveragerc; see issue #193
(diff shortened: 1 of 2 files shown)


embray:
Weird....travis-ci tests mysteriously failed on Python 2.6.  This should have nothing to do with this PR--probably something in the build changed.

I can't reproduce locally...

embray:
The appveyor test on Python 2.6 worked fine too...

embray:
Okay, I was able to reproduce this.  It actually really does fail on Python 2.6 when adding the `--cov-config` option and works fine without it.  Very strange.  Might just disable it on Python 2.6 and move on...

embray:
Interesting...
```
  File "/internal/1/root/home/embray/miniconda/envs/test/lib/python2.6/site.py", line 525, in <module>
    main()
  File "/internal/1/root/home/embray/miniconda/envs/test/lib/python2.6/site.py", line 508, in main
    known_paths = addsitepackages(known_paths)
  File "/internal/1/root/home/embray/miniconda/envs/test/lib/python2.6/site.py", line 288, in addsitepackages
    addsitedir(sitedir, known_paths)
  File "/internal/1/root/home/embray/miniconda/envs/test/lib/python2.6/site.py", line 185, in addsitedir
    addpackage(sitedir, name, known_paths)
  File "/internal/1/root/home/embray/miniconda/envs/test/lib/python2.6/site.py", line 155, in addpackage
    exec line
  File "<string>", line 1, in <module>
  File "<string>", line 4, in <module>
  File "/internal/1/root/home/embray/miniconda/envs/test/lib/python2.6/site-packages/pytest_cov/embed.py", line 58, in init
    auto_data=True)
  File "/internal/1/root/home/embray/miniconda/envs/test/lib/python2.6/site-packages/coverage/control.py", line 138, in __init__
    "Couldn't read '%s' as a config file" % config_file
coverage.misc.CoverageException: Couldn't read 'astropy_helpers/tests/coveragerc' as a config file
```

embray:
Okay, interesting, and subtle.  This has to do with recent updates to pytest-cov that enable it, in a clever but hackish way, to collect coverage data on code run in subprocesses.  However, this hack breaks in a bad way if the path you give to the coveragerc file is a relative path, and your test changes directories before running the subprocess.

The breakage actually occurs very early on in the Python interpreter startup (for the subprocess) in site.py--Python 2.7 and up are able to ignore the exception and resume startup, but Python 2.6's site.py is less flexible and it will crash the whole interpeter startup if an unhandled exception occurs while processing a .pth file.

I'll report this to pytest-cov.  As a workaround, for now, I think I can pass a full path in for the config file.

PR #194, opened 2015-09-30
Fix py.test rootdir on OSX

embray:
This is a followup to #189 needed for things to work properly in OSX, where `tempfile.mkdtemp` does *not* return a real path.

This will also need to be backported to the v1.0.x branch.


embray:
Mysterious appveyor fail on Python 2.6...

embray:
Incidentally, the PR in which I tested this against Astropy, astropy/astropy#4202, passed.  So I went ahead and merged it.  Not sure what the build problems are that I'm having here but I think it's just a build system problem and not a real problem.

embray:
Backported in 61cb890652c6aa46827a202c4a005a81d7dc4c7a, and also ported to astropy/astropy#4020

PR #196, opened 2015-10-03
Fix undefined usage of ‘six’

astrofrog:
@embray - I wasn't sure which module to use to test this since none of the built-in ones have version numbers, correct? Maybe ``setuptools``?


embray:
It's not important--the code is just copied directly over from astropy, and I just missed this reference to the six module.

PR #197, opened 2015-10-20
Stream output from sphinx build even when -w is set

Cadair:
This is a little hacky, but my initial testing says it works under 3.5 and 2.7.

### DIFF
--- a/astropy_helpers/commands/build_sphinx.py
+++ b/astropy_helpers/commands/build_sphinx.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
 import inspect
 import os
 import pkgutil
@@ -170,18 +172,20 @@ class AstropyBuildSphinx(SphinxBuildDoc):
         # prevents a continuous updating at the terminal, but there's no
         # apparent way around this.
         if self.warnings_returncode:
-            proc = subprocess.Popen([sys.executable],
+            proc = subprocess.Popen([sys.executable, '-c', subproccode.encode('utf-8')],
                                     stdin=subprocess.PIPE,
                                     stdout=subprocess.PIPE,
                                     stderr=subprocess.STDOUT)
-            stdo, _ = proc.communicate(subproccode.encode('utf-8'))
-            stdo = stdo.decode('utf-8')
-
-            print(stdo)
-
-            stdolines = stdo.splitlines()
-
-            if 'build succeeded.' in stdolines:
+#            stdo, stde = proc.communicate(subproccode.encode('utf-8'))
+            stdolines = []
+            with proc.stdout:
+                for line in iter(proc.stdout.readline, b''):
+                    line = line.strip(b'\n')
+                    stdolines.append(line)
+                    print(line)
+            proc.poll()
+            #stdolines = stdo.splitlines()
+            if b'build succeeded.' in stdolines:
                 retcode = 0
             else:
                 retcode = 1
--- a/astropy_helpers/commands/build_sphinx.py
+++ b/astropy_helpers/commands/build_sphinx.py
@@ -172,7 +172,7 @@ class AstropyBuildSphinx(SphinxBuildDoc):
         # prevents a continuous updating at the terminal, but there's no
         # apparent way around this.
         if self.warnings_returncode:
-            proc = subprocess.Popen([sys.executable, '-c', subproccode.encode('utf-8')],
+            proc = subprocess.Popen([sys.executable, '-c', subproccode],
                                     stdin=subprocess.PIPE,
                                     stdout=subprocess.PIPE,
                                     stderr=subprocess.STDOUT)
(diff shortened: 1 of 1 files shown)


Cadair:
ping @astrofrog

Cadair:
inspired by: https://stackoverflow.com/questions/2715847/python-read-streaming-input-from-subprocess-communicate/17698359#17698359i

Cadair:
@embray ping

astropy_helpers/commands/build_sphinx.py
```diff
@@ -170,18 +172,22 @@ def run(self):
...
+                    line = line.strip(b'\n')
+                    stdolines.append(line)
+                    print(line.decode('utf-8'))
 
-            stdolines = stdo.splitlines()
+            # Poll to set proc.retcode
+            proc.poll()
 
-            if 'build succeeded.' in stdolines:
+            if b'build succeeded.' in stdolines:
```
embray: One thing we could do to improve this, is it seems there's no reason to save the stdout lines once they've been printed.  Instead set `retcode = 1` by default, and do the check for "build succeeded" in the loop above (and if found, set `retcode = 0`.

Seems fine otherwise--thanks for fixing this.

astropy_helpers/commands/build_sphinx.py
```diff
@@ -170,18 +172,22 @@ def run(self):
...
+            stdolines = []
+            with proc.stdout:
+                for line in iter(proc.stdout.readline, b''):
+                    line = line.strip(b'\n')
+                    stdolines.append(line)
+                    print(line.decode('utf-8'))
 
-            stdolines = stdo.splitlines()
+            # Poll to set proc.retcode
+            proc.poll()
```
eteq: It might be safer to guard this with something like:
```
while proc.poll() is None:
    time.sleep(.5)
```
I've seen travis (and maybe RTD) sit for a while at the end of the process and it's not entirely clear why, so it's conceivable that something weird is going on that might lead to `stdout` getting closed before the process has actually finished.

eteq:
@Cadair - see my comment above in addition to @embrays.  Otherwise it looks good except that appveyor is failing, thinking that the return code is 1.  I wonder if this is due to some platform-dependent behavior with regard to what Popen pipes yield?  If so, my hunch is that the critical point is the change from ``if 'build succeeded.' in stdolines:`` to ``if b'build succeeded.' in stdolines:``.  Might it instead work to do ``if 'build succeeded.' in str(stdolines):``?

astropy_helpers/commands/build_sphinx.py
embray: That's what [`Popen.wait`](https://docs.python.org/2/library/subprocess.html#subprocess.Popen.wait) is for, but otherwise good point in general.

Cadair: I think I avoided that because of the warning in the docs?

embray: It's unlikely to be an issue.  The `iter(proc.stdout.readline, b'')` loop will only, in principle, break once the program has ended (EOF is reached).  You don't have to worry about the stdout buffer filling up because you've already been draining it the whole time.  Using wait here will just make absolutely sure the child process is dead before interrogating it for a return code.

Cadair:
I think this fixes everything.

Cadair:
Any ideas on what is causing this: https://ci.appveyor.com/project/Astropy/astropy-helpers/build/1.0.293/job/cc2d0h0eclmbsj7o#L1211

astropy_helpers/commands/build_sphinx.py
```diff
@@ -170,21 +172,22 @@ def run(self):
         # prevents a continuous updating at the terminal, but there's no
         # apparent way around this.
         if self.warnings_returncode:
-            proc = subprocess.Popen([sys.executable],
+            proc = subprocess.Popen([sys.executable, '-c', subproccode.encode('utf-8')],
```
embray: You're mixing bytes and strings here.  On Python 3 `sys.executable` and `'-c'` are strings, while the result of `.encode('utf-8')` is a bytes object.  In this case you don't need the `.encode('utf-8')`, I think.

Cadair:
yay! It passed!

embray:
Awesome, thanks @Cadair!  I want to test this out a bit real quick, but otherwise I'll probably go ahead and merge.

Cadair:
@embray no problem, thanks.

embray:
Looks great--the next trick will be to capture and output the bare carriage returns during the percent counters, so it doesn't output new lines when lines are overwritten :D  But we can worry about that some other time.  I'm just going to squash your commits a bit before merging.

embray:
I squashed this and added a changelog entry--manually merged into master in ebd45c613d5b772c8ca8765330145740d8f94522

Issue #117, opened 2014-12-18
Sphinx documentation should include module constants that have docstrings


mhvk:
This apparently used to work, but seems to have stopped. See https://github.com/astropy/astropy/pull/3208#issuecomment-67406543 (refiled from https://github.com/astropy/astropy/issues/3226 per @embray's comment).

cdeil:
We also have this issue in Gammapy that things are missing in our docs because of this bug:
https://github.com/gammapy/gammapy/issues/111#issuecomment-71041680

Anyone has the skills / time to investigate and fix this? @eteq @bsipocz ?

astrofrog:
Also an issue for the matplotlib style dictionary in ``astropy.visualization``

mhvk:
@embray - would you have a pointer towards where one could handle this? I am now building a new package (hopefully, affiliated at some point) and have been trying, but failed to do anything useful. I did note that `automodapi` explicitly doesn't make a summary for it. 

Strangely, constants in classes are happily shown with their docstrings.

embray:
I really have no idea.

mhvk:
See #198 for an idea towards a solution. Comments very welcome.

bsipocz:
the reincarnation of #198, astropy/sphinx-automodapi#24,  got merged into ``sphinx-automodapi``, thus I'm closing this.

PR #198, opened 2015-10-21
Allow also non-functions/classes to be included in module API summary

mhvk:
This is a possible solution to #117, in that it adds an option `:include-all-objects` to `automodapi` that allows one to make a summary of objects that are neither classes nor functions. It works by writing an `automodsumm` section with a new `:no-functions-or-classes` option.

I'm not sure about the approach, or what should be default, and would like to have comments before updating tests and documentation.

### DIFF
--- a/astropy_helpers/extern/automodapi/automodapi.py
+++ b/astropy_helpers/extern/automodapi/automodapi.py
@@ -127,7 +127,7 @@ Functions
 """
 
 automod_templ_other = """
-Other variables
+Global variables
 {otherhds}
 
 .. automodsumm:: {modname}
--- a/astropy_helpers/extern/automodapi/automodapi.py
+++ b/astropy_helpers/extern/automodapi/automodapi.py
@@ -9,6 +9,11 @@ module/package is desired in the documentation, use `automodule`_ instead of
 
 It accepts the following options:
 
+    * ``:include-all-objects:``
+        If present, include not just functions and classes, but all objects.
+        This includes variables, for which a possible docstring after the
+        variable definition will be shown.
+
     * ``:no-inheritance-diagram:``
         If present, the inheritance diagram will not be shown even if
         the module/package has classes.
@@ -121,6 +126,15 @@ Functions
     {clsfuncoptions}
 """
 
+automod_templ_other = """
+Other variables
+{otherhds}
+
+.. automodsumm:: {modname}
+    :no-functions-or-classes:
+    {clsfuncoptions}
+"""
+
 automod_templ_inh = """
 Class Inheritance Diagram
 {clsinhsechds}
@@ -209,6 +223,7 @@ def automodapi_replace(sourcestr, app, dotoctree=True, docname=None,
             inhdiag = maindocstr = top_head = True
             hds = '-^'
             allowedpkgnms = []
+            allowothers = False
 
             # look for actual options
             unknownops = []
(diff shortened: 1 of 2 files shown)


Cadair:
ooo this would be useful!

embray:
Thanks for tackling this.  I'm not so sure we need or want a special option for this though.  In the case of #117 for example it used to *just work*, but something changed in Sphinx that broke it, and I never tracked down what changed.

embray:
As you can see this also broke some tests, though it should be easy enough to fix.

astropy_helpers/sphinx/ext/automodapi.py
```diff
@@ -116,6 +121,15 @@
     {clsfuncoptions}
 """
 
+automod_templ_other = """
+Other variables
```
embray: How about just "Global variables"?

mhvk:
> Thanks for tackling this. I'm not so sure we need or want a special option for this though. In the case of #117 for example it used to just work, but something changed in Sphinx that broke it, and I never tracked down what changed.

@embray - given the above, do you think I should pursue this PR, or is it better to try to find out what changed? If I should pursue it, is this the right direction? What should be the default?  (I'm slightly wary since I still don't understand the whole process all that well.)

embray:
I'm not entirely sure either. I didn't write any of this code.  My point is that if we're going to auto-generate documentation for a module that should include any module-level constants as well (at *least* if they have docstrings).  It should do this by default.

embray:
This section in the docs explains things: http://docs.astropy.org/en/stable/development/docguide.html#sphinx-extensions

astropy_helpers/sphinx/ext/automodsumm.py
```diff
@@ -119,6 +119,7 @@ class Automodsumm(BaseAutosummary):
     option_spec = dict(Autosummary.option_spec)
     option_spec['functions-only'] = flag
     option_spec['classes-only'] = flag
+    option_spec['no-functions-or-classes'] = flag
```
embray: Maybe instead of `no-functions-or-classes`, which is awkward, this could just be called `'global-variables-only'`.

embray:
Now I'm *not* entirely sure if this actually used to work or not.  It's been a long time.  I think one thing that did break was display of docstrings for class attributes.  But that's different.

mhvk:
@embray - I agree that the only thing I now for sure used to work was class attributes, but those indeed go through a different code path. I like your suggestion of `global-variables-only` and think it makes most sense to display them by default *if* they have documentation *or* they are in `__all__`.

embray:
Sounds good.

eteq:
@mhvk @embray - :+1: on displaying constants if documented, but a clarification on this proposal: I think it probably is safer to *require* that they be in `__all__`.  My reasoning is that if you don't do that, then modules that do things like ``import numpy as np`` at the top without an `__all__` will be getting the numpy docs included. 

Another option might be to leave on (or maybe turn on?  I can't remember the default for sure...) the machinery that checks to see if the thing getting documented is actually a part of the same module.  But I'm guessing that's more confusing because it might hide stuff that a user actually wanted.

embray:
@eteq I'm definitely on board with only auto-documenting variables in `__all__`

mhvk:
I rebased this long-standing item -- still like the idea. I changed the names according to the suggestions, and hope to look at how to add tests after seeing what failures this gets (it failed in its initial incarnation).

mhvk:
actually, it looks like I should really make a PR to the new `sphinx-automodapi` package, right?

mhvk:
Since it seems this is now in the wrong package, see https://github.com/astropy/sphinx-automodapi/pull/24

If it indeed should be moved there, this one can be closed.

bsipocz:
@mhvk - Indeed, all the sphinx related extensions were moved already, and we only update the bundled version in ``extern`` to match newer ``sphinx-automodapi`` versions. Closing this now, as you've already opened the PR to the upstream repo.

PR #199, opened 2015-10-26
making sure run commands returns 1 if fails

dpshelio:
I've noticed in sunpy/sunpy#1591 that the doctest runs where labelled as successful when they were not.
The problem was that the `proc.returncode` was not being used in `sys.exit()` so we were getting things like this:
```bash
Sphinx Documentation subprocess failed with return code 1

The command "if [[ $TEST_MODE == 'doctest' ]];...." exited with 0.
```
The changes proposed seems to make it work... but I'm not sure if that breaks anything else.

### DIFF
--- a/astropy_helpers/commands/build_sphinx.py
+++ b/astropy_helpers/commands/build_sphinx.py
@@ -216,6 +216,7 @@ class AstropyBuildSphinx(SphinxBuildDoc):
         else:
             log.warn('Sphinx Documentation subprocess failed with return '
                      'code ' + str(proc.returncode))
+            retcode = proc.returncode
 
         if retcode is not None:
             # this is potentially dangerous in that there might be something


embray:
I'm a little confused what this would have to do with doctests, but otherwise this seems right.

dpshelio:
@embray I've found this problem while searching why the doctests were failing but Travis was giving it as good. I don't remember at the moment why it happened in the doctests and not in the sphinx build... I would have to look into it again.

embray:
Apparently the full line in SunPy's .travis.yml is

```
 - if [[ $TEST_MODE == 'doctest' ]]; then python setup.py build_sphinx -b doctest; fi
```

I've never used the Sphinx "doctest" builder, instead just use py.test to run doctests as well.  But okay, this still looks good.

embray:
Not sure why, but Okay, I got the build for this to pass: https://ci.appveyor.com/project/Astropy/astropy-helpers/build/1.0.263  I don't know why there's a separate continuous-integration/appveyor check (distinct from "continueous-integration/appveyor/pr") that doesn't update when I re-run the build.  But oh well this is fine.

embray:
Thanks!

Issue #65, opened 2014-07-30
pip install of all affiliated package is broken with conda?


cdeil:
This is a continuation of https://github.com/astropy/astropy/issues/2759 ... @astrofrog mentioned I should file an issue in the astropy-helpers tracker.

The problem that `pip install` fails for Astropy affiliated packages with an error message about the C compiler, even if the affiliated package is pure Python.

I tried `ccdproc`, `spectral-cube` and `photutils` ... here's the error message for `spectral-cube`:
https://gist.github.com/cdeil/06109508ce883a084d08, 
https://gist.github.com/cdeil/276c3649890b298126ff

This should be easy to reproduce ... install Anaconda Python 3.4 on Ubuntu 12.04 or Mac OS X 10.9, then upgrade to Astropy 0.4, then run

```
pip install ccdproc
```

PR #76, opened 2015-11-20
Compiler check should be skipped for pure-Python packages

embray:
As mentioned by @embray in https://github.com/astropy/astropy-helpers/issues/65#issuecomment-50691291 it would be nice if the compiler check were skipped for pure-Python affiliated packages.

The advantage is that then the user can't run into an installation issue because the compiler check can fails (see the issue mentioned above).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,6 +32,10 @@ astropy-helpers Changelog
   beginning of the build processes (that is, is automatically downloaded via
   setuptools' processing of ``setup_requires``). [#185]
 
+- Moves the ``adjust_compiler`` check into the ``build_ext`` command itself,
+  so it's only used when actually building extension modules.  This also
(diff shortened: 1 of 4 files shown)


embray:
Attached patch that refactors things so that this should no longer be an issue.

If this works this should be accompanied with updates in astropy and package-template.  I'm not 100% I haven't broken the functionality of `adjust_compiler` since I have no great way to test it.  Though the functional tests still seem to work.

(In any case it's not too likely to come up since probably not a lot of people have llvm-gcc on their macs anymore?)  We'll see...

eteq:
@embray - can you clarify exactly *what* changes need to be done in `setup.py` ?

embray:
@eteq Just remove the line that starts with `adjust_compiler...` (the import for it can also be removed).  It's just noise at this point.

PR #127, opened 2015-11-20
Get tests working on py.test 2.6

embray:
Way back in 4d82b0e3f7eda8175aed42a6f399eb1e65fd58ab I disabled running the tests in py.test 2.6 due to some issues in py.test that were causing it to break.  See for example https://bitbucket.org/hpk42/pytest/issue/412/crash-on-bad-stdout#comment-11617176

There are some other bugs I get on py.test 2.6 as well that seem to have more to do with it than with our tests (though it's possible we're exercising some gnarly corner-cases, so I don't blame py.test too much).  This is just a reminder that this should be addressed eventually.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ install:
 
     - conda create --yes -n test python=$PYTHON_VERSION
     - source activate test
-    - conda install --yes pip "pytest<2.6" sphinx cython numpy
+    - conda install --yes pip pytest sphinx cython numpy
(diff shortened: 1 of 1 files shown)


embray:
Looks good~

embray:
Oh, should fix this for appveyor too though....

embray:
Nope.  Looks like appveyor.yml has always just been taking the default pytest version from conda.

PR #205, opened 2015-11-23
Partial workaround for #204

### LINKED ISSUE #204: Argument parsing for build_ext command is generally broken
Over in astropy/astropy#4277 I discovered that when running the setup.py build_ext command with some command-line options, such as

```
$ ./setup.py build_ext --inplace --use-system-wcslib
```

the code that tests whether or not the `--use-system-wcslib` flag was enabled fails.  The command still runs, and uses the `--inplace` flag correctly.  But any code that tries to read the command line flags *prior* to the `setup()` function running (for example [`astropy.wcs.setup_package`](https://github.com/astropy/astropy/blob/master/astropy/wcs/setup_package.py)) will fail.

This problem is specific to the `build_ext` command, but exposes a more general problem with `./setup.py` command-line parsing.  In fact the reason this broke was because of some horrible hacks I had to perform in #185 in order to deal with the subtleties of distutils command-line handling.

I'm going to try to find a new workaround, but I thought I would use this as an opportunity to explain how distutils command-line handling works, and why it's difficult to deal with.

In distutils, each of the individual commands--sub-commands of the `setup.py` script such as `setup.py install`, `setup.py build`, etc. are implemented as *classes*.  These classes have a few class-level attributes on them (namely one called `user_options`) that define what command-line arguments can be accepted by that command.  When it parses the arguments passed to a sub-command, it requires that each of those arguments is explicitly defined on the command class.  Otherwise command-line parsing fails completely.

By default distutils (and its extension setuptools) comes with several commands pre-defined.  However, it is possible to add *new* commands, or replace existing commands with a new command of the same name, by registering new command classes with distutils.  When replacing an existing command it is also common-place to subclass the original implementation of that command in order to add new functionality.  astropy-helpers for example subclasses several of the built-in commands in order to add new functionality (or in some cases work around bugs): https://github.com/astropy/astropy-helpers/tree/master/astropy_helpers/commands

(By far one of the more complicated subclassed commands in astropy-helpers is `build_ext`, which adds significant functionality for making it easier to deal with Cython extensions--we'll get back to that.)

New/replacement commands are registered with distutils by passing an argument called `cmdclass` to the `setup()` function.  `cmdclass` should be a dict mapping the name of the command (e.g. `'install'`) to the class that implements that command (which can have any name).  Because this happens "late" in the `setup.py`--calling `setup()` being the last thing that happens in any `setup.py`--it is impossible to properly parse the `./setup.py` command-line arguments any time before `setup()` is called unless we know ahead of time what command classes *will* be registered with distutils.

This is why astropy-helpers has a function called [`register_commands`](https://github.com/astropy/astropy-helpers/blob/v1.1b1/astropy_helpers/setup_helpers.py#L269) that is called early in the setup.py.  This returns a dictionary of all commands that astropy-helpers adds/extends.  This is created early on so that astropy-helpers can then use distutils' command-line parsing in order for individual sub-packages to make decisions based on what command-line arguments the user provided (for example, the `astropy.wcs` package has to be configured differently depending on whether the `--use-system-wcslib` argument was provided).  This works okay, but is pretty kludgy.  It also makes it hard for affiliated packages to further add/extend commands.  They have to provide their own wrapper around `register_commands` in order for it to really work with astropy-helpers.

Where this totally breaks down, however, is in the current implementation of Astropy's `build_ext` command, which builds extension modules.  It actually has two different modes of operation:  One when Cython is installed, and one when it isn't.  When building Astropy from a source distribution, it actually doesn't need Cython at all, because we included the Cython-generated sources in the distribution.  astropy-helpers still provides a little additional functionality, so it still subclasses the original `build_ext` command, but this isn't a big deal.

However, for building from a git checkout, Cython is needed in order to generate the C sources.  Cython itself comes with its own extension to the `build_ext` command in `Cython.Distutils.build_ext`.  It includes several new command-line options specific to Cython, and ensures that any Cython sources that have been updated are reinterpreted and new C sources generated (before the C sources themselves are built).

astropy-helpers actually determines at run-time whether or not Cython should be used by the `build_ext` command, and uses the appropriate subclass (either setuptools.build_ext or Cython.build_ext) depending.  This worked fine (if not kludgy) until #185 in which I tried to be too clever.  The problem #185 tried to solve is that I wanted to make it possible to load Cython automatically even when a user does not have Cython already installed, by way of the `setup_requires` mechanism.  `setup_requires` is an argument that is passed to `setup()` which supplies the names of packages that should be installed at build-time in order for the build to succeed.  So in principle it's possible to temporarily install Cython as soon as it's needed by the build process.

There is a long-acknowledged catch-22 to the `setup_requires` feature which limits its usefulness: because requirements are passed in via `setup()`, the requirements are only processed as soon as the `setup()` function is called.  This means that your `setup.py` can't use any of the `setup_requires` packages *before* the `setup()` is run.  #185 tried to make it so that Cython can be passed to `setup_requires` by making it so that the build_ext command is only created late into the setup process (after `setup()` has been called), so that it can determine whether or not Cython is available *after* `setup_requires` has been processed.

This seemed to work.  But it means that anywhere before `setup()` is called it can't reliably report what command-line options are supported by the `build_ext` command.  Therefore any earlier attempt to read the command-line options for that command fails.

All of this speaks to several of the long-acknowledged problems with distutils' architecture.  d2to1 works around several of these problems rather effectively, and it is my long-term desire to switch astropy over to using d2to1, or a newer descendant to it.  In the meantime I think I will handle this just by rolling back #185 *for now*.

embray:
This partially fixes #204 by providing at least the standard `build_ext` options by default, whether or not Cython is installed (incidentally, something I forgot is that setuptools actually does subclass its build_ext option from the Cython build_ext, but it doesn't provide this "late subclassing" of Cython when Cython is provided via setup_requires :/).

This means the problem in #204 is fixed so long as the user does not supply cython-specific options to the build_ext command.  If they do, and Cython is not installed, then functions like [`astropy_helpers.distutils_helpers.get_distutils_option`](https://github.com/astropy/astropy-helpers/blob/v1.1b1/astropy_helpers/distutils_helpers.py#L56) will still fail, because distutils command-line parsing will fail due to the presence of unrecognized options.

A better long-term solution will still be to write our own command-line parsing for `./setup.py` that can allow unrecognized options :/

This workaround will be fine for the vast majority of likely real-world cases though so I'm happy with this for v1.1, with the new command-line parsing something we can address later if needed.

### DIFF
--- a/astropy_helpers/commands/_dummy.py
+++ b/astropy_helpers/commands/_dummy.py
@@ -76,6 +76,6 @@ if sys.version_info[0] < 3:
     """))
 else:
     exec(dedent("""
-        class _DummyCommand(Command, metaclass=_DummyCommandMeta):
+        class _DummyCommand(Command, object, metaclass=_DummyCommandMeta):
             pass
     """))
--- a/astropy_helpers/commands/_dummy.py
+++ b/astropy_helpers/commands/_dummy.py
@@ -1,0 +1,81 @@
+"""
+Provides a base class for a 'dummy' setup.py command that has no functionality
+(probably due to a missing requirement).  This dummy command can raise an
+exception when it is run, explaining to the user what dependencies must be met
+to use this command.
+
+The reason this is at all tricky is that we want the command to be able to
+provide this message even when the user passes arguments to the command.  If we
+don't know ahead of time what arguments the command can take, this is
+difficult, because distutils does not allow unknown arguments to be passed to a
+setup.py command.  This hacks around that restriction to provide a useful error
+message even when a user passes arguments to the dummy implementation of a
+command.
+
+Use this like:
+
+    try:
+        from some_dependency import SetupCommand
+    except ImportError:
+        from ._dummy import _DummyCommand
+
+        class SetupCommand(_DummyCommand):
+            description = \
+                'Implementation of SetupCommand from some_dependency; '
+                'some_dependency must be installed to run this command'
+
+            # This is the message that will be raised when a user tries to
+            # run this command--define it as a class attribute.
+            error_msg = \
+                "The 'setup_command' command requires the some_dependency "
+                "package to be installed and importable."
+"""
+
+import sys
+from setuptools import Command
+from distutils.errors import DistutilsArgError
+from textwrap import dedent
+
+
+class _DummyCommandMeta(type):
+    """
+    Causes an exception to be raised on accessing attributes of a command class
+    so that if ``./setup.py command_name`` is run with additional command-line
+    options we can provide a useful error message instead of the default that
+    tells users the options are unrecognized.
+    """
+
+    def __init__(cls, name, bases, members):
+        if bases == (Command, object):
+            # This is the _DummyCommand base class, presumably
+            return
+
+        if not hasattr(cls, 'description'):
+            raise TypeError(
+                "_DummyCommand subclass must have a 'description' "
+                "attribute.")
+
+        if not hasattr(cls, 'error_msg'):
+            raise TypeError(
+                "_DummyCommand subclass must have an 'error_msg' "
+                "attribute.")
+
+    def __getattribute__(cls, attr):
+        if attr in ('description', 'error_msg'):
+            # Allow cls.description to work so that `./setup.py
+            # --help-commands` still works
+            return super(_DummyCommandMeta, cls).__getattribute__(attr)
+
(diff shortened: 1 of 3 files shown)


bsipocz:
This was merged to master before 1.1 no matter what the release scripts say (was part of a nested  master merge, thus not picked up).

PR #172, opened 2015-11-24
Method summary table in HTML docs is empty

embray:
The method summary table in the HTML docs is empty:

Example:  http://astropy.readthedocs.org/en/latest/api/astropy.table.Table.html
screen shot 2015-07-15 at 11 29 23

I think this affects all Astropy classes and affiliated packages (at least it does in Gammapy).

@eteq @embray Any idea how to fix that?

### DIFF
--- a/astropy_helpers/sphinx/ext/astropyautosummary.py
+++ b/astropy_helpers/sphinx/ext/astropyautosummary.py
@@ -111,3 +111,12 @@ def setup(app):
     if LooseVersion(sphinx.__version__) < LooseVersion('1.2.0'):
         # this replaces the default autosummary with the astropy one
         app.add_directive('autosummary', AstropyAutosummary)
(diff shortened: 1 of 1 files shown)


embray:
The attached patch fixes it.  Since this is just a workaround to an upstream bug I haven't attached a test, but I have confirmed manually that this fixes it.

eteq:
@embray - oops, yeah, you're right, I was confused about the actual problem here. 

Curiously enough, I independently fixed the exact same problem just yesterday in a much hackier way (astropy/halotools#245).  This is definitely a better solution!  Looks good to me assuming the tests pass.

embray:
Oops--too bad about the wasted duplicate effort.

bsipocz:
This was merged to master before 1.1.0 no matter what the release scripts say (was part of a nested  master merge, thus not picked up).

PR #214, opened 2016-01-13
Python3 links not working

mhvk:
In https://github.com/astropy/astropy/issues/4465, I found that using `python3 setup.py build_sphinx`, links to python3 `SyntaxWarning` and `RuntimeWarning` were not working.  I can solve this locally by replacing the local storage file for python3 listed in https://github.com/astropy/astropy-helpers/blob/master/astropy_helpers/sphinx/conf.py#L46 with `None`. Does that make sense?


mhvk:
@bsipocz - I turned this into a PR, but am not sure this is the right way to go about it; if not, please feel free to make an alternative PR.

bsipocz:
@mhvk - Can you confirm that this works with python2? I fear that removing the local hack inventory will brake those links that are listed in it at the moment.

mhvk:
@bsipocz - sadly, it doesn't seem this simple solution suffices. Building astropy documentation on python2, I seem to get only warnings for a missing reference to `bytes`, while on python3, I get warnings about missing references to `unicode`, plus missing references to `file`, `cPickle` and various `exceptions.*`. I think some of these are actual python2 assumptions in the documentation; the `bytes/unicode` thing is hopefully easier to fix.

embray:
I don't think preserving inter-sphinx references to all Python builtins is deeply important in all cases.  I like this solution, and for any cases where it doesn't work add nitpick exceptions.

bsipocz:
See #216 for a solution that works for both py2 ad py3. Since most affiliates uses python2 I think that we should keep the py2 compatibility of the helpers for a little longer (even if switching to py3 sphinx builds in astropy master).

bsipocz:
This can be closed as the superseding #216 is merged.

PR #215, opened 2016-01-23
Add ways to include inherited members into generated class docs

### LINKED ISSUE #109: Inherited members always ignored in automodsumm
The sphinx extension ``automodsumm`` always ignores inherited class members when producing an auto summary. This is different from the default behavior of ``autosummary``. I'd like to add a configuration variable to ``automodsumm`` to be able to revert this (perhaps ``automodsumm_inherited_members``), but I wanted to bring it up here, as the behavior differs from ``autosummary``, which I find to be interesting / a possible bug.

hamogu:
This adds a global sphinx option `automodsumm_inherited_members`.
The default is `False` matching the previous behavior.
However, if set to `True` inherited members of classes will be included
into the generated API documentation.

Note that I added no tests for this behavior. There are no tests for the exiting
behavior either and adding tests for those kind of things is a lot of work.
However, I generated docs and checked by hand that it works as exepcted.

closes #109
[skip ci]

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,10 @@ astropy-helpers Changelog
 1.1.3 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Added sphinx configuration value ``automodsumm_inherited_members``.
+  If ``True`` this will include members that are inherited from a base
+  class in the generated API docs. Defaults to ``False`` which matches
+  the previous behavior. [#215]
 
 
 1.1.2 (2016-03-9)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,16 +4,16 @@ astropy-helpers Changelog
 1.2 (unreleased)
 ----------------
 
-- Nothing changed yet.
+- Added sphinx configuration value ``automodsumm_inherited_members``.
+  If ``True`` this will include members that are inherited from a base
+  class in the generated API docs. Defaults to ``False`` which matches
+  the previous behavior. [#215]
 
 
 1.1.3 (unreleased)
 ------------------
 
-- Added sphinx configuration value ``automodsumm_inherited_members``.
-  If ``True`` this will include members that are inherited from a base
-  class in the generated API docs. Defaults to ``False`` which matches
-  the previous behavior. [#215]
+- Nothing changed yet.
 
 
 1.1.2 (2016-03-9)
--- a/astropy_helpers/sphinx/ext/automodapi.py
+++ b/astropy_helpers/sphinx/ext/automodapi.py
@@ -52,12 +52,12 @@ It accepts the following options:
         given, only objects that are actually in a subpackage of the package
         currently being documented are included.
 
-    * ``:inherited-members: True | False``
+    * ``:inherited-members:`` / ``:no-inherited-members:``
         The global sphinx configuration option
         ``automodsumm_inherited_members`` decides if members that a class
         inherits from a base class are included in the generated
-        documentation. The option ``:inherited-members:`` allows to
-        overrride this global setting.
+        documentation. The option ``:inherited-members:`` or ``:no-inherited-members:``
+        allows the user to overrride the global setting.
 
 
 This extension also adds two sphinx configuration options:
--- a/astropy_helpers/sphinx/ext/automodapi.py
+++ b/astropy_helpers/sphinx/ext/automodapi.py
@@ -52,6 +52,14 @@ It accepts the following options:
         given, only objects that are actually in a subpackage of the package
         currently being documented are included.
 
+    * ``:inherited-members: True | False``
+        The global sphinx configuration option
+        ``automodsumm_inherited_members`` decides if members that a class
+        inherits from a base class are included in the generated
+        documentation. The option ``:inherited-members:`` allows to
+        overrride this global setting.
+
+
 This extension also adds two sphinx configuration options:
 
 * ``automodapi_toctreedirnm``
@@ -204,6 +212,7 @@ def automodapi_replace(sourcestr, app, dotoctree=True, docname=None,
 
             # look for actual options
             unknownops = []
+            inherited_members = None
             for opname, args in _automodapiargsrex.findall(spl[grp * 3 + 2]):
                 if opname == 'skip':
                     toskip.append(args.strip())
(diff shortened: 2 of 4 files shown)


hamogu:
I would like a configuration option to switch this on or off for individual blocks, too, but I have not done that yet.
In fact, I would argue that astropy should use `automodsumm_inherited_members=True` and then switch it off for those objects that derive from numpy arrays (nddata, quantity).

hamogu:
@embray : I believe you are the best person to review this.
Added options `:inherited-members:` and `:no-inherited-members:` to override the global `automodsumm_inherited_members` flag that I introduced in the last commit. It's not pretty to have two options here, but I experimented with other ways and found them too confusing, e.g. have a single options `:do-inherited-members-opposite-to-global-setting:` would allow all the flexibility but the effect of this flag would depend on the setting of the global option - too confusing.

hamogu:
Unfortunately, I did not find a way to implement this without copying code from sphinx...

hamogu:
@eteq @embray : Please let me know if / when you have time to review this so that I can rebase it (I just don't want to rebase it every time a conflict comes up unless I know that somebody will look at it in the near future).

hamogu:
Forgot the say: @astrofrog Please let me know if / when you have time to review this so that I can rebase it (I just don't want to rebase it every time a conflict comes up unless I know that somebody will look at it in the near future).

eteq:
@hamogu - sorry about the silence here.  I'll try to review this within the next few days.

I have a vague memory of some trick I managed to figure out to work around having to hack `find_autosummary_in_lines`.  Will also try to dig that up in case it applies here...

CHANGES.rst
```diff
@@ -4,7 +4,10 @@ astropy-helpers Changelog
 1.1.2 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Added sphinx configuration value ``automodsumm_inherited_members``.
```
eteq: This should be in the 1.2 section, as it's adding a new feature.  (although see my comment in astropy/astropy-helpers#216 about the idea of doing a 1.2 release of the helpers prior to the 1.2 release of astropy)

astropy_helpers/sphinx/ext/automodapi.py
```diff
@@ -52,6 +52,14 @@
         given, only objects that are actually in a subpackage of the package
         currently being documented are included.
 
+    * ``:inherited-members: True | False``
+        The global sphinx configuration option
+        ``automodsumm_inherited_members`` decides if members that a class
+        inherits from a base class are included in the generated
+        documentation. The option ``:inherited-members:`` allows to
+        overrride this global setting.
```
eteq: nitpick: "allows to overrride" -> "allows overriding" or "allows the user to override" or similar.

astropy_helpers/sphinx/ext/automodapi.py
```diff
@@ -52,6 +52,14 @@
         given, only objects that are actually in a subpackage of the package
         currently being documented are included.
 
+    * ``:inherited-members: True | False``
```
eteq: The fact that it's a global configuration option shouldn't be documented here.  This part of the documentation is for options to the *directive*, not configuration settings.  You should instead add a ``automodsumm_inherited_members`` bullet to the list immediately after this (the one right after "This extension also adds two sphinx configuration options"), and the documentation here should just document  ``:inherited-members:`` (with no ``True | False``, as the True/False status is set by its presence or absence)

eteq: Oh, and it should probably be stated explicitly that you should use ``:no-inherited-members:`` to negate a default set to True.  Maybe just have the bullet point be ```* ``:inherited-members:`` / ``:no-inherited-members:`` ```

astropy_helpers/sphinx/ext/utils.py
```diff
@@ -63,3 +67,111 @@ def find_mod_objs(modname, onlylocals=False):
         objs = [e for i, e in enumerate(objs) if valids[i]]
 
     return localnames, fqnames, objs
+
+
+def find_autosummary_in_lines(lines, module=None, filename=None):
```
eteq: I'd suggest renaming this to something like `find_autosummary_in_lines_for_automodsumm`.  Just to make sure someone else looking at it doesn't think it's for more general use or a bug-fix of sphinx or something.

eteq:
OK, I finished reviewing, @hamogu.  I'm also not happy with having to copy  `find_autosummary_in_lines`, and thought a bit about ways to work around it, but ended up concluding as you did that it's probably best just to accept this, as the probability of it changing is low.

astropy_helpers/sphinx/ext/automodapi.py
hamogu: I don't see this as documentation for `automodsumm_inherited_memebers`. As you say, that is documented elsewhere (below). If you have an idea how I can word this and propeprly explain what it does without mentioning that `automodsumm_inherited memebers` is (a) the global option (so you know what you are overriding) and (b) what it does (so you know what effect it will have when you set the override) please let me know.

I don't see how I can avoid any dublication of info about automordsumm_inherited_memeber.

hamogu:
Pushed changes. @eteq: Let me know if I could convince you about mentioning `automodsumm_inherited_members` here: https://github.com/astropy/astropy-helpers/pull/215#discussion-diff-55459704R55 or if you have a better suggestion.

hamogu:
I don't understand the Travis failures. I can only assume that something went wrong with the rebase and the version string is screwed up or something. Any ideas?

eteq:
@hamogu - the failures are unrelated (see #222).

And what you have here is just what I had in mind, so we're all set in the issue from https://github.com/astropy/astropy-helpers/pull/215#discussion-diff-55459704R55 .  I have a couple minor wording quibbles but will just fix those directly and merge into master.  Thanks!

bsipocz:
@hamogu @eteq - This is just the thing I was looking for astroquery in https://github.com/astropy/astroquery/issues/677. However it looks a little bit overkill to also include the special ``__method__``-s. Is that intentional or something we can fix?

e.g. ``login()`` and ``authenticated()`` nicely shows up compared to the current docs [0], but ``__call__()`` too, and we don't want that.

screen shot 2016-04-28 at 16 07 36

http://astroquery.readthedocs.io/en/latest/api/astroquery.eso.EsoClass.html#astroquery.eso.EsoClass

hamogu:
There is some special treatment of __init__ and __call__ here:
 https://github.com/astropy/astropy-helpers/pull/215/files#diff-c95430c65faa667beb1feea7abfb3fd6R564
I did not write those lines (I only broke them to fit the line lengths limit) and I did not think through them in detail.

From my perspective that's something that can be fixed - but I rarely comment `__init__` and `__call` in my code anyway; I usually use the class description for that. That's probably why I did not notice.

PR #219, opened 2016-03-04
log.warn is deprecated, so use log.warning instead

keflavich:
At least, that's how I understand this message:

    DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead

### DIFF
--- a/ah_bootstrap.py
+++ b/ah_bootstrap.py
@@ -361,7 +361,7 @@ class _Bootstrapper(object):
         dist = self._directory_import()
 
         if dist is None:
-            log.warn(
+            log.warning(
                 'The requested path {0!r} for importing {1} does not '
                 'exist, or does not contain a copy of the {1} '
                 'package.'.format(self.path, PACKAGE_NAME))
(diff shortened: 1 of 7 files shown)


bsipocz:
:+1:

eteq:
@keflavich - hmm, the test failures suggest that some of these's aren't regular python Loggers, and hence don't have `warning`?  Not sure which, though...

bsipocz:
@eteq - You're right, it's a different log ``from distutils import log``. 
I had some deja vu when saw this PR, as remembering that I tried to get rid of similar deprecations, and wondered how this could have slipped through :)

eteq:
So are they *all* the `distutils` log, in which case we should close this, or are some of these regular log objects that should be `warning`?

keflavich:
ah, my bad, I didn't check where `log` was coming from; I was doing global search-and-replaces in other projects where this was all `logging.warn`, which is deprecated.  I'll check tomorrow & maybe just close.

eteq:
OK, great, thanks @keflavich

keflavich:
Can confirm, all `log`s are from distutils:
```
$ ack 'import log'
ah_bootstrap.py
131:from distutils import log

astropy_helpers/commands/build_ext.py
10:from distutils import log, ccompiler, sysconfig

astropy_helpers/commands/build_sphinx.py
12:from distutils import log

astropy_helpers/setup_helpers.py
19:from distutils import log

astropy_helpers/tests/__init__.py
84:    from distutils import log

astropy_helpers/version_helpers.py
29:from distutils import log

ez_setup.py
25:from distutils import log
```

PR #220, opened 2016-03-10
Included licenses in MANIFEST.in

eteq:
While working on astropy/astropy#4682, I noticed that the licenses directory for the helpers was left out of the ``MANIFEST.in``.  We should include this for consistency, if nothing else.


astrofrog:
:+1:

eteq:
Alright, simple enough so will go ahead and merrge

PR #224, opened 2016-04-13
Fixing extension warnings raised by sphinx 1.4

bsipocz:
There are two new warnings coming from the new 1.4 Sphinx. This causes travis failures for the sphinx builds for packages using astropy-helpers. The proposed workaround https://github.com/astropy/ci-helpers/pull/78 is to set the version to a previous release, however on the long term, it's better to deal with them here.

This PR addresses the second warning, as I'm not yet sure how to fix the first one.

```
WARNING: while setting up extension astropy_helpers.sphinx.ext.autodoc_enhancements: directive 'autoattribute' is already registered, it will be overridden
WARNING: sphinx.ext.pngmath has been deprecated. Please use sphinx.ext.imgmath instead.
```

[edit: I've cancelled the appveyor build as it's irrelevant and there was a queue for other packages]

Issue #222, opened 2016-03-11
Figure out why the travis tests are failing on the "eva" version name


eteq:
About half of the helpers' tests are currently failing due to a few instances of ``SystemExit: error: Invalid distribution name or version syntax: -eva--1.0.dev1``.  E.g., https://travis-ci.org/astropy/astropy-helpers/builds/115207264

It doesn't seem to be affecting the functionality, so maybe it's a change in how version parsing works?  @astrofrog, I think you were going to look at this?  Or maybe @embray has some insight?

bsipocz:
@eteq - This is down to the new v20.x setuptools. Tests pass with older versions. 
Even though we test against the dev version, the latest test that run and passed was against setuptools==19.8.dev0 in https://travis-ci.org/astropy/astropy-helpers/jobs/107074804.

So while this doesn't solve the issue I suggest to limit the setuptools version number until someone figures the solution out, thus we still see other failures genuine to a given PR. (And also add an extra one or two builds with the "stable" setuptools to the allowed failures as a reminder that something needs to be fixed).

We may also consider limiting the sphinx version number until both parts of #224 is solved.

Should I open a PR with these?

eteq:
Alright, this is ameliorated by #225.  

But @bsipocz, do you have any idea what the underlying problem is? Or @astrofrog or @embray?

bsipocz:
Nope, sadly I don't have a clue.

I think I would do it with the brute force approach and start diffing setuptools 19.8 and 20.1. But I won't have time for it any time soon (as both the helpers and setuptools seem too complex, so I know I would run into stupid deadends first).

astrofrog:
I'm git bisecting right now.

astrofrog:
The first failing commit in setuptools is https://github.com/pypa/setuptools/commit/a718819d2849196e902808301c9a95724510c5c1

astrofrog:
I think it might just be that the test package we use starts and ends with an underscore? (``_eva_``)

astrofrog:
Confirmed: changing ``_eva_`` to ``eva`` fixes it.

astrofrog:
Now the question is whether this is a setuptools bug, but I don't have time to investigate that right now...

PR #225, opened 2016-04-15
Limiting versions of setuptools and sphinx

bsipocz:
This is to deal temporarily with the latest travis failures that are not affecting functionality.

See details in #222 and #224.

ping @eteq

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,8 @@ env:
 matrix:
     include:
         - os: linux
+          env: PYTHON_VERSION=2.7 SETUPTOOLS_VERSION='<20' DEBUG=True
+        - os: linux
           env: PYTHON_VERSION=2.7 SETUPTOOLS_VERSION=dev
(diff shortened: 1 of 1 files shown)


bsipocz:
I think the failure is due to an unrelated coveralls glitch, restarting that build should make it pass.

eteq:
@bsipocz - yep, I think you're right - re-started travis and if it passes this is good to merge.

The Appveyor failures are more mysterious.... but it looks like it's been present for a while so I won't worry about that right now.

eteq:
I really don't get why coveralls is unhappy.  I'm 99% sure its unrelated though, so I'm going to merge this now and we'll see if its consistent over the next few builds.

bsipocz:
> The Appveyor failures are more mysterious.... but it looks like it's been present for a while so I won't worry about that right now.

Seems somewhat related, at least it went wrong at the same time.

bsipocz:
Yep, same problem there, setuptools got upgraded to 20.x.

PR #226, opened 2016-04-15
Cleanup PR

bsipocz:
This PR does a few low prio cleanup tasks:

 * Opts in the travis container for the linux builds.
 * Removes the locally patched version of the sphinx extension viewcode
 * Fixes PEP8, except long lines and except the few third party files hosted in sphinx/ext

I think the only controversial point is the viewcode removal. The patch got included 1.5 years ago into 1.3b1. I think it's a long enough time, and we can remove our local version.

Note that this fails, but when built on top of #225 the tests pass (https://travis-ci.org/bsipocz/astropy-helpers/builds/123386933).


astropy_helpers/setup_helpers.py
```diff
@@ -273,9 +269,9 @@ def run_command_hooks(cmd_obj, hook_kind):
         if isinstance(hook, str):
             try:
                 hook_obj = resolve_name(hook)
-            except ImportError as exc:
+            except ImportError:
                 raise DistutilsModuleError(
-                        'cannot find hook {0}: {1}'.format(hook, err))
+                    'cannot find hook {0}: {1}'.format(hook, err))
```
embray: This looks like a bug.  My guess is pylint or whatever warned that `exc` is not used, but actually it is on this line but the variable was accidentally renamed `err` here.  Should change back to `as exc` and change `err` to `exc` here.

bsipocz: Thanks, indeed.

eteq:
Hmm, it's a fair question what the oldest version of sphinx we want to support is.  I think older sphinxes are less of a priority because it's sort of a "dev-oriented" thing to need to build the sphinx docs.

Note, though, @bsipocz, that I think it makes sense to update the `need_sphinx` in `conf.py` to 1.3 if you're removing the viewcode.

bsipocz:
@eteq - changed ``need_sphinx`` to 1.3, and rebased. I guess this will need a changelog entry, but not sure which one; 1.2?

eteq:
Hmm, yeah, now that you ask, I suppose upgrading the required sphinx is something we wouldn't want in a bugfix.  So I'd say 1.2

eteq:
@bsipocz - actually, @astrofrog raised a good point that this affects the affiliated packages, so you should ask on astropy-affiliated-maintainers if anyone minds having to upgrade to 1.3 .  If everyone there is OK with it I'd say it's ok to include this for helpers v1.2

bsipocz:
That's a good point, I'll send out an e-mail.

bsipocz:
@eteq - There was no reaction to the e-mail about this, so I assume everyone is OK with upgrading to sphinx >=1.3. 
There was quite many changes in the helpers recently, so I rebased the PR.

astrofrog:
@bsipocz - can you rebase?

bsipocz:
@astrofrog - Done

astrofrog:
I think AppVeyor is broken for unrelated reasons, so I'll go ahead and merge and debug that in another issue.

bsipocz:
I was just about to open an issue about the appveyor failures, as said on one of the other PR it's around for a long time.

Issue #227, opened 2016-04-15
Fix the remaining sphinx 1.4 warning


eteq:
As discussed in #224, sphinx 1.4 has caused two new warnings to appear.  One was fixed in #224, but this one remains:
```
WARNING: while setting up extension astropy_helpers.sphinx.ext.autodoc_enhancements: directive 'autoattribute' is already registered, it will be overridden
```

And from https://github.com/astropy/astropy-helpers/pull/224#issuecomment-210034280:
>  The first warning is coming from this https://github.com/astropy/astropy-helpers/blob/master/astropy_helpers/sphinx/ext/autodoc_enhancements.py#L75
>
> I tried to catch the warning there and ignore it, but it didn't work. Getting the bug workaround into upstream sphinx may be the best long term solution.

bsipocz:
btw @embray made a PR to sphinx with the patch here: https://github.com/sphinx-doc/sphinx/pull/1843

eteq:
And I'm noticing you've suggested an in-sphinx fix in sphinx-doc/sphinx#2451

bsipocz:
Yes, either way would work. I tried to implement the latter but over-complicated it, and still didn't manage to get it work.

PR #229, opened 2016-04-19
Work around sphinx 1.4 warning

eteq:
This PR fixes the remaining warning in sphinx 1.4 (see #227). It's a terribly hacky fix, but it's the only way I see to supress the warning. 

Hopefully sphinx-doc/sphinx#2451 will eventually get addressed, and then we can instead suppress the warning in a more reasonable way.

cc @bsipocz

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ env:
     - PYTHON_VERSION=3.5
   global:
     - SETUPTOOLS_VERSION='<20'
-    - SPHINX_VERSION=stable
     - CONDA_DEPENDENCIES="setuptools pytest sphinx cython numpy"
     - PIP_DEPENDENCIES="coveralls pytest-cov"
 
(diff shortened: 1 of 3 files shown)


bsipocz:
@eteq - There is some progress on sphinx-doc/sphinx#2451, but it doesn't seem working for me. Maybe link that sphinx issue before the try/finally as a reminder.

So :+1: merging this now, and once a working patch is out in an upstream release, we can remove this hack.

eteq:
Sounds good - I will add a comment pointing to that issue and then merge this.  Thanks @bsipocz !

eteq:
Alright, merging this.  Will create a new issue to replace this once sphinx 1.4.2 is out.

bsipocz:
Just to confirm it here sphinx-doc/sphinx#2451 works as expected. 
(I just messed up the pulling of the latest dev version on top of my locally hacked one in the afternoon.)

bsipocz:
@astrofrog - This is another one for backportation (however at some point should be removed, but sphinx 1.4.2 is not yet ready on conda just via pip).

PR #232, opened 2016-04-20
Adding the suppress_warnings config item ...

bsipocz:
…  to ignore the warning raised by overriding of directives, as we mean to do that. 

This can step into the place of #229 once Sphinx 1.4.2 is out.

@eteq - Do you prefer to merge this now, or should I also include the removal of #229 in here, and leave it open until the new sphinx is out?

### DIFF
--- a/astropy_helpers/sphinx/conf.py
+++ b/astropy_helpers/sphinx/conf.py
@@ -81,6 +81,12 @@ rst_epilog = """
 .. _Astropy: http://astropy.org
 """
 
+# A list of warning types to suppress arbitrary warning messages. We mean to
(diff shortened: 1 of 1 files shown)


eteq:
@bsipocz - sorry I dropped the ball here - I just tried it on sphinx 1.4.2 (which has come out in the interim) and this fixes it.  I'm going to do a manual merge of this that adds another commit to turn off #229 if the sphinx version in use is >=1.4.2 .

eteq:
There @astrobot.  Are you happy?  Jerk.

bsipocz:
@astrofrog - and this may be also up for grabs for a backportation although the alternative in #229 does the same job but in a hacky way (and we need the hack for sphinx 1.4 and 1.4.1).

PR #235, opened 2016-05-16
Add "generated" to list of directories build_sphinx -l removes

eteq:
Does just what it says.  This is intended to go with astropy/astropy#4924, and together they will make it so that ``python setup.py build_sphinx -l`` properly cleans out all the build products of the new sphinx gallery extension that's used in astropy.

It also creates a de facto expectation that other sphinx extensions that generate something should put their products in ``generated`` (or ``api``, although that's sort of a special case).

### DIFF
--- a/astropy_helpers/commands/build_sphinx.py
+++ b/astropy_helpers/commands/build_sphinx.py
@@ -67,7 +67,8 @@ class AstropyBuildSphinx(SphinxBuildDoc):
     def finalize_options(self):
         #Clear out previous sphinx builds, if requested
         if self.clean_docs:
-            dirstorm = [os.path.join(self.source_dir, 'api')]
+            dirstorm = [os.path.join(self.source_dir, 'api'),
+                        os.path.join(self.source_dir, 'generated')]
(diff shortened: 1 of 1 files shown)


bsipocz:
Are you going to open a PR for package-template, too to update the makefile?

eteq:
No, but I should!  Thanks for the reminder, @bsipocz - see astropy/package-template#171

eteq:
@Cadair - I tried to figure out a way to have this access the ``conf.py`` in sphinx to make the "generated" definition a bit more flexible... But it's just not possible because ``finalize_options`` always gets run before any sphinx context is created.

So the question is: are you ok with this as-is, or would you like me to add a check for a file along the lines of ``docs/api/dontdelete``?  Or a command-line option to specify all the generated doc directories? (which you could put in sunpy's ``setup.cfg``)

eteq:
I'm going to go ahead and merge this to move things along... but if you think something like what I said in the last comment is a good idea, @Cadair, it would be easy to implement as a follow-on PR.

Issue #237, opened 2016-05-30
Intersphinx: collections (python2) -> collections.abc (python3)


MSeifert04:
Is it possible to setup an intersphinx mapping that links the python3 ``collections.abc`` as well as the python2 ``collections`` correctly?

For example ``collections.Mapping`` (py2) is now ``collections.abc.Mapping`` (py3). But the not-abcs like ``collections.OrderedDict`` remained in the ``collections`` namespace.

bsipocz:
Ideally it should be fine if the code works using this as ``collections.Mapping`` for both versions. Did you see sphinx warnings when building the docs?

bsipocz:
OK, now I see it locally. I think this kind of things should have dealt with in the official python interpshinx, but since it isn't I'll add it to our local one. Thanks @MSeifert04 for spotting it.

astrofrog:
@bsipocz - can this be closed?

bsipocz:
@astrofrog - absolutely

PR #238, opened 2016-05-31
Adding collections.Mapping to the local python3 inventory

bsipocz:
As pointed out in #237 this causes a problem with the python3 sphinx builds (while it still can be imported from the old namespace).

This is also needed for astropy/astropy#4556 to pass.


astrofrog:
@bsipocz - any idea what happened here?

bsipocz:
@astrofrog - This is weird, the isn't any sphinx test on appveyor so this shouldn't affect anything there.

bsipocz:
Ohh, now I got some vague memories back. The appveyor builds are failing for a long time, but in the last helpers spring we didn't have time to fix it and forgot about it since (at least I forgot about it).

astrofrog:
The AppVeyor failure is unrelated, so merging this.

Issue #243, opened 2016-07-05
automodapi - :no-main-docstr: breaks links


saimn:
When I use something like

```
.. automodapi:: mpdaf.tools
   :no-main-docstr:
```

I then get a lot of warnings about wrong references, and the generated doc does not have the links to functions/methods. Example for one function:

```
/home/simon/dev/mpdaf/doc/tools.rst:52: WARNING: failed to import copy_header
...
/home/simon/dev/mpdaf/doc/tools.rst:52: WARNING: toctree references unknown document u'api/copy_header'
...
/home/simon/dev/mpdaf/doc/api/mpdaf.tools.copy_header.rst:: WARNING: document isn't included in any toctree
...
```

It means that the generated autostuff code does not have correct references. Adding `:no-main-docstr:` has only one effect in the generated RST, it removes a line :

```
.. automodule:: {modname}
```

But then the generated code also has a `.. currentmodule:` directive, so everything should be ok. Any idea of what could be wrong here ?

bsipocz:
automodapi got moved to https://github.com/astropy/sphinx-automodapi, so please reopen there is this is still relevant.

saimn:
For the record, I think it was closed by https://github.com/astropy/sphinx-automodapi/commit/b4ea01384a05a865ea03361f115aed7ad227631f but I didn't test.

PR #244, opened 2016-07-05
Fix encoding error with automodapi_writereprocessed and py2

saimn:
`automodapi_writereprocessed` was not working on Python 2

### DIFF
--- a/astropy_helpers/sphinx/ext/automodapi.py
+++ b/astropy_helpers/sphinx/ext/automodapi.py
@@ -91,6 +91,7 @@ This extension also adds two sphinx configuration options:
 # actually built.
 
 import inspect
+import io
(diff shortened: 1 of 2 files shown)


pllim:
Is this to fix #243?

saimn:
@pllim - No, it unrelated.

saimn:
Up ! (it is really a minor fix, useful mainly for debugging purpose.)

bsipocz:
This looks good to me, and already got the thumbs up from @embray, so I'm merging. Tagged as no changelog-needed, but I'm half sure about that, maybe one would be useful.

saimn:
Thanks @bsipocz ! No changelog is fine for me.

PR #249, opened 2016-08-28
Add support for package that lies in a subdirectory

saimn:
When the root package lies in a subdirectory, like `lib` or `src`, it is already possible to specify a `srcdir` parameter for some functions, but a few things are missing, and a few paths are using `packagename` instead of `package_dir`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ astropy-helpers Changelog
 2.0 (unreleased)
 ----------------
 
+- Add support for package that lies in a subdirectory. [#249]
 
 1.3.2 (unreleased)
 ------------------
--- a/astropy_helpers/commands/build_ext.py
+++ b/astropy_helpers/commands/build_ext.py
@@ -255,12 +255,14 @@ def generate_build_ext_command(packagename, release):
 
             extensions = self.distribution.ext_modules
             if extensions:
+                build_py = self.get_finalized_command('build_py')
+                package_dir = build_py.get_package_dir(packagename)
                 src_path = os.path.relpath(
                     os.path.join(os.path.dirname(__file__), 'src'))
                 shutil.copy(os.path.join(src_path, 'compiler.c'),
-                            os.path.join(self.package_name, '_compiler.c'))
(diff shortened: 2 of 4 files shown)


embray:
Looks good to me.

I'm generally against such source layouts for a couple reasons:
1) It's a bit more of a pain to do in-source imports.  One has to add `src/` to Python-path somehow, or explicitly use `./setup.py develop`.  It's nicer to just be able to cd to the source directory and import from there.
2) A lot of Python-based tools (like astropy-helpers, apparently!) just don't support that very well.

All that said, there are some packages where 1) doesn't necessarily apply--especially projects that are not pure-Python.  And also I'm not one to force a particular source layout on anyone either, so if this change made that work and you've tested it I'm all for it.  (In fact a test would be kind of nice to have for this, but I won't demand it--it's hard to write such a test, especially such that it hits all the relevant code).

saimn:
Thanks for reviewing this Erik !
I used to prefer the simple layout, but I changed my mind after reading some articles from ionelmc ([1](https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure) [2](https://blog.ionelmc.ro/2014/06/25/python-packaging-pitfalls/) and maybe others), which somewhat convinced me that being able to import the package from the source directory (your point 1) is a bad idea.
But for astropy_helpers it seems to do some magic to make sure that the package is properly build (and installed ?) before testing, so it should not be an issue.

Also I'm trying to convert a package which has a `lib/` sub-directory, so if I can avoid to break my other development branches and the same for the others developers, it would be nice. :)

bsipocz:
@saimn - Since this get the thumbs up from @embray, I'm happy to merge after a rebase and green CI.

saimn:
@bsipocz - Ok, I will do the rebase and some testing since this is quite old (it could be a good opportunity to try again this package migration to astropy_helpers !).

saimn:
@bsipocz - Done!

bsipocz:
Brilliant. Could you add a changelog, too to 2.0?

bsipocz:
Thanks @saimn!

saimn:
Thanks for merging ;)

Issue #250, opened 2016-09-06
AstropyBuildSphinx TypeError


cdeil:
I started a new package from the package-template and got this error:

```
$ python setup.py $SETUP_CMD
/home/travis/miniconda/envs/test/lib/python2.7/site-packages/matplotlib/font_manager.py:273: UserWarning: Matplotlib is building the font cache using fc-list. This may take a moment.
  warnings.warn('Matplotlib is building the font cache using fc-list. This may take a moment.')
Freezing version number to healpix/version.py
running build_sphinx
creating /home/travis/build/cdeil/healpix/docs/_build
creating /home/travis/build/cdeil/healpix/docs/_build/doctrees
creating /home/travis/build/cdeil/healpix/docs/_build/html
/home/travis/build/cdeil/healpix/astropy_helpers/astropy_helpers/commands/build_sphinx.py:244: AstropyDeprecationWarning: The "build_sphinx" command is now deprecated. Use"build_docs" instead.
  '"build_docs" instead.', AstropyDeprecationWarning)
Traceback (most recent call last):
  File "setup.py", line 119, in <module>
    **package_info
  File "/home/travis/miniconda/envs/test/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/home/travis/miniconda/envs/test/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/home/travis/miniconda/envs/test/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/home/travis/build/cdeil/healpix/astropy_helpers/astropy_helpers/commands/build_sphinx.py", line 245, in run
    super(AstropyBuildSphinx, self).run()
TypeError: super() argument 1 must be type, not classobj
```

See https://github.com/cdeil/healpix/pull/2 where @astrofrog mentioned that I should file an issue here.

Some facts:
- Error only on Python 2, Python 3 works fine
- Error only for the deprecated `build_sphinx`, not the to-be used `build_docs` (`build_sphinx` is which is still the default in the package-template `.travis.yml` at this time)

I don't see the error when just running `python setup.py build_sphinx` on the package-template, but I de see it in this new project which basically is just a copy of the package-template files.

If someone has the energy / motivation to try and reproduce / fix this, you should be able to do it quickly by checking out the right commit corresponding to the travis-ci build linked to above and running `python2 setup.py build_sphinx`.

[...]

PR #246, opened 2016-07-07
Deprecate build_sphinx

adrn:
This is a fix for #245. This switches the inheritance order of `AstropyBuildDocs` and `AstropyBuildSphinx` and adds a deprecation warning whenever `AstropyBuildSphinx` is run.

PR #251, opened 2016-09-06
Fix AstropyBuildSphinx

saimn:
These classes inherit from distutils' `Command` which is an old-style class so
`super` fails (on Python 2 only).
Ref #250, cc @cdeil

### DIFF
--- a/astropy_helpers/commands/build_sphinx.py
+++ b/astropy_helpers/commands/build_sphinx.py
@@ -242,4 +242,4 @@ class AstropyBuildSphinx(AstropyBuildDocs): # pragma: no cover
         warnings.warn(
             'The "build_sphinx" command is now deprecated. Use'
             '"build_docs" instead.', AstropyDeprecationWarning)
-        super(AstropyBuildSphinx, self).run()
+        AstropyBuildDocs.run(self)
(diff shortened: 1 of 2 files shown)


embray:
I feel like there has to be something else going on here, because as far as I know this has normally worked on Python 2 (otherwise the astropy doc builds would not have been working for quite some time).  But :+1: if this fixes it.

embray:
Oh I see, this was just to the deprecated `build_sphinx`--that explains it.   Weren't there plans to remove that soon?

saimn:
Yep, the `super` call was added recently (#246), when `build_sphinx` was deprecated. It seems this change has not yet reached the template (and maybe affiliated packages) so maybe too early to remove the command.

cdeil:
@saimn - Thanks!

> It seems this change has not yet reached the template (and maybe affiliated packages) so maybe too early to remove the command.

Definitely. This fix should go in, or `build_sphinx` should be removed.
I've had others run into this same issue of calling `build_sphinx` out of habit and being confused.

@astrofrog @bsipocz - Can you merge this PR or is there some other maintainer that can?

saimn:
Up !

bsipocz:
Thanks @saimn for the fix.

PR #261, opened 2016-10-27
Force the use of Cython's old build_ext command

saimn:
A new build_ext command was added in Cython 0.25 (https://github.com/cython/cython/pull/1456) but it breaks Astropy (I could not find why, see https://github.com/astropy/astropy/issues/5437).

Cython 0.25.1 was released to revert this change:
https://github.com/cython/cython/commit/4ecdd3e4e5984011fe93af6c4ccaf6d67a88f6f5

This commit ensures that we always use the old build_ext, whatever the Cython version is. At least until we find why the new build_ext command does not work.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,10 @@ astropy-helpers Changelog
 - ``build_sphinx`` has been deprecated in favor of the ``build_docs`` command.
   [#246]
 
+- Force the use of Cython's old ``build_ext`` command. A new ``build_ext``
+  command was added in Cython 0.25, but it does not work with astropy-helpers
+  currently.  [#261]
+
 
 1.2 (2016-06-18)
 ----------------
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -43,5 +43,5 @@ install:
 build: false
 
 test_script:
-  - "%CMD_IN_ENV% py.test"
+  - "%CMD_IN_ENV% py.test astropy_helpers"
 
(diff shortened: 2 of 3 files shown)


eteq:
Thanks @saimn... This seemed fine to me by eye, but the tests are failing in a variety of convoluted ways... any idea why?

saimn:
@eteq - No idea ... Locally tests pass on Py2, and on Py3 I had to use `PYTHONDONTWRITEBYTECODE=1` to avoid "sandbox violation" errors with the `__pycache__` directories, but I also get these errors on master.

pllim:
That `pyx` failure is no more. So is this still relevant?

MSeifert04:
@pllim I think so, because if I understand it correctly they just reverted to issue a deprecation warning:

```
C:\...\lib\site-packages\Cython\Distutils\old_build_ext.py:30: UserWarning: Cython.Distutils.old_build_ext does not properly handle dependencies and is deprecated.
  "Cython.Distutils.old_build_ext does not properly handle dependencies "
```

saimn:
@pllim - I think it is, as using the new build_ext breaks Astropy it is better to make sure that we always use the old one (until we find what is not working with the new). 
However I have no idea why tests are failing ...

saimn:
Re-running the tests to make sure it was not something broken on Travis/Conda side.

saimn:
And Travis now passes :open_mouth: 
Appveyor fails because it tries to run tests for ci-helpers (`ERROR collecting ci-helpers/test_env.py`). Will fix that in `appveyor.yml`

bsipocz:
I think this will need a changelog entry. Could you add one @saimn, and then I can merge?

saimn:
@bsipocz - Rebased and added changelog.

PR #266, opened 2016-11-24
Updating bundled numpydoc version to v0.6

bsipocz:
@astrofrog @eteq - Do we need to backport these kinf of updates? What about changelog, I assume that could be useful?

### DIFF
--- a/astropy_helpers/sphinx/ext/docscrape.py
+++ b/astropy_helpers/sphinx/ext/docscrape.py
@@ -24,10 +24,10 @@ class Reader(object):
            String with lines separated by '\n'.
 
         """
-        if isinstance(data,list):
+        if isinstance(data, list):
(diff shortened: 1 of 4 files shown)


pllim:
I thought Sphinx uses `napoleon` extension now instead of `numpydoc`?

bsipocz:
Afaik we haven't switched yet (e.g. https://github.com/astropy/astropy/issues/3974), and nevertheless I'm not sure we can that easily remove numpydoc from the helpers without deprecating it first.

bsipocz:
Merging this now, but I'm opening and issue here, too to remind us to switch to ``napoleon`` at some point.

bsipocz:
I think this doesn't need a changelog entry, but please confirm @astrofrog.

PR #273, opened 2016-12-03
Skip registering build_docs if sphinx missing some of its dependencies.

### LINKED ISSUE #228: Install fails if Sphinx dependency imagesize is missing
I have Sphinx 1.4 installed, but not it's dependency [imagesize](https://pypi.python.org/pypi/imagesize) that they added in 1.4 (so yes, you could say I have a broken Sphinx install).

This prevents me from installing Astropy or any affiliated package!?

```
$ python2 setup.py install --user
  File "setup.py", line 57, in <module>
    cmdclassd = register_commands(PACKAGENAME, VERSION, RELEASE)
  File "/Users/deil/code/package-template/astropy_helpers/astropy_helpers/setup_helpers.py", line 142, in register_commands
    from .commands.build_sphinx import AstropyBuildSphinx, AstropyBuildDocs
  File "/Users/deil/code/package-template/astropy_helpers/astropy_helpers/commands/build_sphinx.py", line 16, in <module>
    from sphinx.setup_command import BuildDoc as SphinxBuildDoc
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/sphinx/setup_command.py", line 23, in <module>
    from sphinx.application import Sphinx
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/sphinx/application.py", line 38, in <module>
    from sphinx.environment import BuildEnvironment
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/sphinx/environment.py", line 42, in <module>
    from sphinx.util.images import guess_mimetype
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/sphinx/util/images.py", line 13, in <module>
    import imagesize
ImportError: No module named imagesize
```

Can this be made more robust?
A missing dependency for Sphinx shouldn't prevent Astropy installation, right?

bsipocz:
This is to fix #228 

Testing this is not trivial, and probably is unneccesary. Locally I've removed ``imagesize`` that reproduced the error reported in #228, and with this fix everything is working again as expected.

One question though: @cdeil would you prefer to receive a warning about the missing dependency/sphinx, or can it stay as it was for the missing sphinx: simple passing in the try/expect.

### DIFF
--- a/astropy_helpers/setup_helpers.py
+++ b/astropy_helpers/setup_helpers.py
@@ -48,6 +48,7 @@ _module_state = {
 
 try:
     import sphinx
+    from .commands.build_sphinx import AstropyBuildSphinx, AstropyBuildDocs
     _module_state['have_sphinx'] = True
 except ValueError as e:
     # This can occur deep in the bowels of Sphinx's imports by way of docutils
@@ -136,9 +137,7 @@ def register_commands(package, version, release, srcdir='.'):
     if _module_state['registered_commands'] is not None:
         return _module_state['registered_commands']
 
-    if _module_state['have_sphinx']:
-        from .commands.build_sphinx import AstropyBuildSphinx, AstropyBuildDocs
-    else:
+    if not _module_state['have_sphinx']:
         AstropyBuildSphinx = AstropyBuildDocs = FakeBuildSphinx
 
     _module_state['registered_commands'] = registered_commands = {
--- a/astropy_helpers/setup_helpers.py
+++ b/astropy_helpers/setup_helpers.py
@@ -48,7 +48,6 @@ _module_state = {
 
 try:
     import sphinx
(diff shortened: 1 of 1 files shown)


bsipocz:
Very well, so while this fixes the error when tried in an affiliated package install, but make the helpers tests fail at many point.

cdeil:
> One question though: @cdeil would you prefer to receive a warning about the missing dependency/sphinx, or can it stay as it was for the missing sphinx: simple passing in the try/expect.

Ideally Sphinx import would only be tried if `python setup.py build_docs` is issued.
If this is not possible with how the machine works, then I'd expect this:
- For `python setup.py install` no message about missing or broken Sphinx
- For `python setup.py build_docs` an error if Sphinx is missing or broken.

OK?

PR #264, opened 2016-12-05
Link to Github repo from PyPI

cdeil:
From https://pypi.python.org/pypi/astropy-helpers/1.2 there's currently no link to https://github.com/astropy/astropy-helpers

And the other way around, the README doesn't link to PyPI.

I think it's useful to have that, to be able to quickly go back and forth.

I can make a PR if this is OK with the maintainers.
If yes, OK if I change the URL in setup.py (see https://github.com/astropy/astropy-helpers/blob/master/setup.py#L32) to point to https://github.com/astropy/astropy-helpers, or would you prefer to only have it in the text description?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,12 +1,16 @@
 astropy-helpers
 ===============
 
+* Stable versions: https://pypi.org/project/astropy-helpers/
+* Development version, issue tracker: https://github.com/astropy/astropy-helpers
+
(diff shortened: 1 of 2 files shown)


cdeil:
I've attached a commit here.
@bsipocz - Please review.

bsipocz:
Thanks @cdeil. I'm merging now as this is definitely an improvement.

PR #277, opened 2016-12-12
Address pytest deprecations

### LINKED ISSUE #262: [pytest] change in setup.cfg
My travis builds were spitting out warnings about the [pytest] section being deprecated.  I switched this to [tool:pytest] locally and that stopped the warnings.  

I don't know much about this, so figured i'd bring it up to those in the know here first - but i'm happy to issue a quick pull request with the changes if this is something needed.

bsipocz:
Closes #262 

I could only get the other warning go away by removing the version limitation of pytest-cov, hopefully the issue with >=2.3 pytest-cov also went away, but that couldn't be tested with running travis on my fork. Adding milestone 1.3.1 as this is not critical for v1.3.

```
WC1 None pytest_funcarg__cov: declaring fixtures using "pytest_funcarg__" prefix is deprecated and scheduled to be removed in pytest 4.0.  Please remove the prefix and use the @pytest.fixture decorator instead.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ env:
     - PYTHON_VERSION=3.5
   global:
     - CONDA_DEPENDENCIES="setuptools pytest sphinx cython numpy"
-    - PIP_DEPENDENCIES="coveralls pytest-cov==2.2"
+    - PIP_DEPENDENCIES="coveralls pytest-cov"
 
 matrix:
     include:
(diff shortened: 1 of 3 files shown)


bsipocz:
Unfortunately the -60% reported coverage decrease from #246 is still around with coverage 2.4, so I'm afraid we can't avoid to look into the details much longer.

bsipocz:
@eteq - I've added you to review this, basically to decide whether it's a good enough solution to list the temp dirs in coveragerc. It still not solves the question why they ended up being included, but I think is a better solution than pinning the version number.

bsipocz:
After off GH discussion with @eteq, and confirming that this doesn't break affiliates using the astropy shipped pytest this will go in v1.3.

bsipocz:
This was merged to master as well as 1.3 no matter what the release scripts say (was part of a nested backport merge, thus not picked up).

Issue #270, opened 2016-12-03
Switch to use sphinx.ext.napoleon rather than numpydoc


bsipocz:
It was raised in e.g. astropy/astropy#3974 that we should aim to switch to use the sphinx included napoleon rather than our numpydoc. However the transition will need a little bit of work, and also a deprecation period for our shipped numpydoc as affiliated packages use it.

astrofrog:
Are we sure we want to do this? Is numpydoc deprecated or are these now competing efforts?

pllim:
I switched over for my other projects. Never had problem. It is just changing the extension setting. One less dependency to worry about.

astrofrog:
What is the minimum required version of sphinx for napoleon?

pllim:
The doc says "new in 1.3"

bsipocz:
I'm not sure we need to do this, but it was already suggested for the core.
It may be good if napoleon is a bit stricter about the syntax, but I don't
have anything against numpydoc

eteq:
I know that I tried switching to napoleon once as an experiment (hacking with @Cadair, I belive?) and we got a bunch of sphinx warnings that were hard to track down.  But that *might* have been on an earlier version of napoleon or numpydocs?  Was probably ~6 months ago at least that I tried this...

Cadair:
I recently did this swap for SunPy, it caused about 7 warnings, so there was some tweaking. This might obviously be worse for Astropy.

Even if the core package can not switch immediately I would be in favour of changing it in the package-template for new affiliated packages.

bsipocz:
While I opened this issue following the astropy one, the answer to [@astrofrog's question above](https://github.com/astropy/astropy-helpers/issues/270#issuecomment-267353361) is not answered, and quick googling is not helping either. What are the pros and cons of using napoleon over numpydoc?

astrofrog:
I don't feel strongly about this btw, but I can see the benefit of having one less dependency (since napoleon is built into sphinx)

Cadair:
I think they are not competing efforts, my main motivation for change would be one less dep / no need to maintain the astropy fork of numpydoc.

astrofrog:
Moved this issue to https://github.com/astropy/sphinx-astropy/issues/5

PR #278, opened 2016-12-15
Move numpydoc and sphinx-automodapi to astropy_helpers.extern

astrofrog:
This moves the numpydoc (and associated files) and automodapi extensions to astropy_helpers.extern and updates ``conf.py`` accordingly. At the moment this should therefore preserve backward-compatibility for astropy and affiliated packages, but not for packages that added the sphinx extensions manually to ``conf.py`` (I can try and do a search later to see if such cases exist).

The ``astropy_helpers/sphinx/ext`` folder is now a lot tidier and it's clearer which things are developed by us vs externally (i.e. numpydoc had quite a few files mixed in with ours).

Note that I've also removed the tests for the bundled modules because those tests may not always work given the updated module path, and there's no point testing things here that are being tested already in the respective repos.

(@bsipocz - ignore the comment I made before about numpydoc, I made a mistake when copying files)


astrofrog:
I need to look into the automodapi-related failures, I think it's something to do with the path to the API docs directory

pllim:
While you are at it, can you just implement #270 and forget about numpydoc altogether?

astrofrog:
@pllim - no time to do that now so I'd rather keep the two separate

astrofrog:
@pllim - just to clarify, I mean that based on other discussions I've seen, this can create new warnings etc and it's a time consuming process to make sure it all works fine. So I'd rather leave that for another time.

bsipocz:
I can't review this before Sunday. As far as I remember the latest release
eas uaed in #266, and am quite surprised to hear about the differences.

astrofrog:
@bsipocz - there are no differences, ignore my original comment (was doing something silly when copying files).

astropy_helpers/sphinx/conf.py
```diff
@@ -135,15 +135,11 @@ def check_sphinx_version(expected_version):
...
-    'astropy_helpers.sphinx.ext.astropyautosummary',
-    'astropy_helpers.sphinx.ext.autodoc_enhancements',
-    'astropy_helpers.sphinx.ext.automodsumm',
-    'astropy_helpers.sphinx.ext.automodapi',
+    'astropy_helpers.extern.numpydoc',
+    'astropy_helpers.extern.automodapi.automodapi',
     'astropy_helpers.sphinx.ext.tocdepthfix',
     'astropy_helpers.sphinx.ext.doctest',
-    'astropy_helpers.sphinx.ext.changelog_links',
-    'astropy_helpers.sphinx.ext.smart_resolver']
```
bsipocz: Do we not need this any more? It has been moved into automodapi, but we don't do a blank inclusions of everything that is in there, do we?

astropy_helpers/sphinx/conf.py
```diff
@@ -135,15 +135,11 @@ def check_sphinx_version(expected_version):
     'sphinx.ext.coverage',
     'sphinx.ext.inheritance_diagram',
     'sphinx.ext.viewcode',
-    'astropy_helpers.sphinx.ext.numpydoc',
-    'astropy_helpers.sphinx.ext.astropyautosummary',
-    'astropy_helpers.sphinx.ext.autodoc_enhancements',
-    'astropy_helpers.sphinx.ext.automodsumm',
```
bsipocz: While ``automodsumm`` is getting included via ``automodapi`` and ``autodoc_enhancements`` via ``automodsum``, I think it's preferable to list them here to be more explicit.

astropy_helpers/sphinx/conf.py
```diff
@@ -135,15 +135,11 @@ def check_sphinx_version(expected_version):
     'sphinx.ext.coverage',
     'sphinx.ext.inheritance_diagram',
     'sphinx.ext.viewcode',
-    'astropy_helpers.sphinx.ext.numpydoc',
-    'astropy_helpers.sphinx.ext.astropyautosummary',
```
bsipocz: 👍 to finally remove it. I'm not sure whether we need to mention this in the changelog as a removed deprecation.

astropy_helpers/sphinx/conf.py
astrofrog: The ``smart_resolver`` extension is not currently included by default when using ``automodapi``, so we need to include it explicitly.

astropy_helpers/extern/numpydoc/plot_directive.py
```diff
@@ -0,0 +1,642 @@
+"""
```
astrofrog: Ok I removed this and four other files removed in that PR in 026a53a584b30cb5717986cb2580201bd668d9d3

astropy_helpers/sphinx/conf.py
astrofrog: The sphinx-automodapi extension officially only recommends adding ``automodapi`` and optionally ``smart_resolver``, so in that sense it's 'public API' to not include automodsumm and  autodoc_enhancements. But I'm happy to add this back if you really would prefer to see them here.

bsipocz: I didn't know about this recommendation, so ignore my comment.

astropy_helpers/sphinx/conf.py
bsipocz: then it should be listed here, shouldn't it?

astrofrog: Oops I read this the wrong way around :)

astropy_helpers/extern/automodapi/automodapi.py
```diff
@@ -1,13 +1,5 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """
-This sphinx extension adds a tools to simplify generating the API
-documentation for Astropy packages and affiliated packages.
-
-.. _automodapi:
-
-========================
-automodapi directive
```
eteq: Why was this removed?  It seems like the header is useful in the generated docs...?

astropy_helpers/extern/automodapi/automodapi.py
```diff
@@ -381,8 +383,13 @@ def process_automodapi(app, docname, source):
...
 
 def setup(app):
-    # need automodsumm for automodapi
-    app.setup_extension('astropy_helpers.sphinx.ext.automodsumm')
+
+    app.setup_extension('sphinx.ext.autosummary')
+
+    # Note: we use __name__ here instead of just writing the module name in
+    #       case this extension is bundled into another package
+    from . import automodsumm
```
eteq: This is a bit confusing: doesn't this mean it's *always* going to be the bundled one, regardless of whether another copy is installed?  Or is that the *intended* behavior, and a later change would then make an "external" copy possible?

astropy_helpers/extern/automodapi/automodsumm.py
```diff
@@ -1,17 +1,5 @@
...
-This sphinx extension adds two directives for summarizing the public
-members of a module or package.
-
-These directives are primarily for use with the `automodapi`_ extension,
-but can be used independently.
-
-.. _automodsumm:
-
-=======================
-automodsumm directive
```
eteq: ditto as above on removing the header

astropy_helpers/extern/automodapi/automodapi.py
astrofrog: The headings are no longer needed in the generated docs: http://sphinx-automodapi.readthedocs.io/en/latest/automodapi.html

astropy_helpers/extern/automodapi/automodapi.py
astrofrog: The intended behavior is that it always uses the automodsumm that lives in the same directory relative to the automodapi file, regardless of whether we are using the standalone sphinx-automodapi or (like here) bundling it in a package. The idea here is to have code that works for both the standalone and the bundled version.

astrofrog:
@eteq - all the changes to automodapi are actually changes that were made in automodapi since it was split off into a separate package. There are a number of bug fixes (that I identified because I built up an extensive test suite in the standalone package) and the headings are changed because the docs for sphinx-automodapi don't need the headings (as seen e.g. [here](http://sphinx-automodapi.readthedocs.io/en/latest/automodapi.html))

astrofrog:
> One thing I think that needs to be added, though: A clear explanation in the docs about what's going on here. I.e., in the documentation section of the development docs, perhaps a section that specifies how things now work: that some of the extensions are maintained by astropy but in separate repos. Along similar lines, it would be useful to add a note somewhere that clearly says that there are tests of all of this, but they are in the stand-alone repos (which I agree makes sense, but anyone just looking at the astropy one should be clearly redirected so they can run those tests if need be).

I agree - we don't have docs for astropy-helpers though, but I think the right place to put this is in the README for astropy-helpers?

astropy_helpers/extern/automodapi/automodapi.py
eteq: Ohh, I gotcha, so this is only for use *within*.  All is fine then.

eteq:
Ah, ok, I understand better now that this is both the bugfixes *and* the move at the same time. The inline comments should be good then.

> I agree - we don't have proper docs for astropy-helpers though, but I think the right place to put this is in the README for astropy-helpers?

Yes, that probably makes the most sense.  It *could* also go in the docstrings for the sphinx extensions... I.e. ``astropy_helpers/sphinx/ext/__init__.py``.  But it'll probably be more obvious for now to put it in the README.  Or even a mention of it in both?  Either way works for me.

astrofrog:
@eteq - I've added some documentation in https://github.com/astropy/astropy-helpers/pull/278/commits/d98b1b17cc8a3b8f56d4f23a94c98423f06306bf - does this seem ok?

eteq:
Yep, that looks great!  Approving, and merging!

bsipocz:
@eteq - are you adding a changelog manually or shall we leave this as is, so next time releasing I can try out the scripts?

PR #283, opened 2016-12-23
Running cron build for only one version

bsipocz:
I think it should be enough to run only one of the python versions for the cron testing (that currently runs every day), what do you think @astrofrog?


astrofrog:
Sounds good!

bsipocz:
@astrofrog - Should we do something similar for astropy? E.g. no need to run with the old numpy/python and OSX builds.

PR #285, opened 2017-01-23
Update automodapi.py

hamogu:


pllim:
Wasn't this moved to https://github.com/astropy/sphinx-automodapi ?

bsipocz:
@hamogu - Thanks for this fix. However I think we try to follow the policy to bring over the released version of bundled, external packages in ``extern`` rather than patching them locally.

hamogu:
OK, I've also put in the fix upsteam.

PR #289, opened 2017-02-22
Fix a bug with build_doc -l (clean)

alexrudy:
Cleaning must be done after the super-finalize_options call.

Test added.


bsipocz:
Thanks @alexrudy, this indeed does the trick.

bsipocz:
@astrofrog - Could you change astrobot's access here, too? I don't have access to the settings page.

Issue #234, opened 2016-05-13
Move the sphinx extensions in to a seperate installable package


Cadair:
I often find myself going to recommend `automodapi` to people but then realise that they would need to pick up the astropy-helpers submodule and all the stuff to really use them, is there any reason why moving them to a seperate package would be unwelcome?

`automodapi` really is lovely and it seems a shame to keep it hidden away in here.

Cadair:
same goes for the theme while we are at it...

bsipocz:
(well, you can use the same argument for basically everything in astropy-helpers, e.g. I love how the testing is set up, and the extra doctest decorators).

Cadair:
but not everything, the actual setup.py machinery needs to be in a submodule, and that's cool. 

Isn't the test runner in astropy now anyway?

Maybe the doctest stuff should go somewhere else as well?! I don't know much about it.

Cadair:
This is an example of the theme as a standalone package: https://github.com/Cadair/astropy-sphinx-theme

astrofrog:
@eteq and I had a chat about this and we agree it'd be worth trying this.

We should start with automodapi as an example. I think it makes sense to try and create a package ``sphinx-automodapi`` or something like that that includes:

* astropyautosummary (actually do we still need that or is that obsolete?)
* automodapi
* automodsumm
* smart_resolver (since it's basically related to the API docs)
* autodoc_enhancements maybe? (since that seems related?)

and we should include the relevant tests.

Would anyone be interested in doing the work of setting up that package? (@Cadair or @bsipocz or both?) If so I can create a repo in the astropy organization. Once we have that in place, I can test it out with a couple of non-astro packages I have to see if it works, and we can then make a decision whether to go ahead with this.

Note that we should also add sphinx docs in the process to describe how to use this sphinx extension.

pllim:
c/c @justincely since he was just playing with a similar idea recently.

bsipocz:
@astrofrog - in general I'm interesed in this so keep me it the loop, but
can't commit to do much work in the next few weeks as I need to finish up a
few other things.

saimn:
:+1: for this. I use automodapi for a package without astropy-helpers, and having a pip-installable package would make things easier (I had to modify a few things, like the relative imports).

    astropyautosummary (actually do we still need that or is that obsolete?)

[...]

PR #290, opened 2017-02-28
Add a path helper for the theme

Cadair:
This is just to make it a little easier to use astropy_helpers just for the sphinx theme.

### DIFF
--- a/astropy_helpers/sphinx/__init__.py
+++ b/astropy_helpers/sphinx/__init__.py
@@ -6,3 +6,11 @@ the Astropy documentation format. Note that some sphinx extensions which are
 bundled as-is (numpydoc and sphinx-automodapi) are included in
 astropy_helpers.extern rather than astropy_helpers.sphinx.ext.
 """
+
+import os
+
+
+def get_html_theme_path():
+    """Return list of HTML theme paths."""
+    cur_dir = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'themes')
+    return [cur_dir]


embray:
I think it would be best to move the Astropy Sphinx theme into a separate package.  The problem with astropy-helpers is that it was originally focused on just the Astropy project itself.  But since it's part of the package template, and now applies to so many other packages, I think it should really move toward focusing on installation problems more than anything else (*one* of the reasons for its existence in the first place was to be able to address installation problems more deftly without having to make whole new Astropy releases).

Cadair:
@embray I agree, but I very much lost that discussion.

In an ideal world it would be very nice if `astropy_helpers` were to become not astropy specific, I think the biggest blocker to that idea is the maintenance burden.

embray:
I added a comment in #234.  I think it makes a lot of sense for the sphinx theme to live on its own.

Cadair:
In other news, merge this?!

bsipocz:
@Cadair - merging this is fine I think, but I'm still not sold on the theme part as is. 

What about an unconference/hacking together on an astropy customized version of sphinx_rtd_theme? That would be robust enough to make others happy (https://github.com/astropy/pyregion/pull/96#issuecomment-237558583) and also would have the astropy branding and possible extras that we put together here. @eteq @astrofrog?

embray:
Problem with merging this is then it becomes harder to *remove* in the future, when it might not be relevant if the theme were made available as a separate package ;)

astrofrog:
See https://github.com/astropy/astropy-helpers/issues/338 - TL;DR we agreed to move the theme out into a separate package. @Cadair - can I close this?

Issue #288, opened 2017-02-22
build_docs swallows errors b/c of changed code in Sphinx


alexrudy:
When there is an error in the Sphinx build, fails in astropy-helpers' build_docs command fails with 
```
Traceback (most recent call last):
  File "<stdin>", line 39, in <module>
NameError: name 'self' is not defined
Sphinx Documentation subprocess failed with return code 1
```

This is because astropy injects the run method of build_docs into a subprocess, but doesn't inject the command object itself. The command object (``self``) is referenced in [line 190 of sphinx/setup_command.py](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/setup_command.py#L190). The code generated for the subprocess is

```
from __future__ import print_function
from sphinx.setup_command import *

os.chdir(u'/Users/alexrudy/Dropbox/Library/Development/Astronomy/Zeeko/docs')
sys.path.insert(0, u'/Users/alexrudy/Dropbox/Library/Development/Astronomy/Zeeko/build/lib.macosx-10.11-x86_64-2.7')
sys.path.insert(0, '/Users/alexrudy/Dropbox/Library/Development/Astronomy/Zeeko/astropy_helpers')

if not color_terminal():
    nocolor()
if not 3:
    status_stream = StringIO()
else:
    status_stream = sys.stdout
confoverrides = {}
if '':
    confoverrides['project'] = ''
if '':
    confoverrides['version'] = ''
if '':
    confoverrides['release'] = ''
if '':
    confoverrides['today'] = ''
if '':
    confoverrides['copyright'] = ''

app = None
try:
    with docutils_namespace():
        app = Sphinx(u'./docs', u'./docs',
                     u'./build/sphinx/html', u'./build/sphinx/doctrees',
                     'html', confoverrides, status_stream,
                     freshenv=False,
                     warningiserror=False)
        app.build(force_all=False)
        if app.statuscode:
            raise DistutilsExecError(
                'caused by %s builder.' % app.builder.name)
except Exception as exc:
    handle_exception(app, self, exc, sys.stderr)
    if not False:
        raise SystemExit(1)

if False:
    src = app.config.master_doc + app.builder.out_suffix
    dst = app.builder.get_outfilename('index')
    os.symlink(src, dst)

```

I'm not sure what to do about this. Maybe pickle the command object? That seems bad, and swallowing the error is really frustrating.

saimn:
[...]

PR #292, opened 2017-03-04
Fix build_sphinx error catcher

alexrudy:
This is the fix suggested by @saimn for issue #288 

It works, but it is still really fragile to upstream changes in Sphinx. I think a more sophisticated solution could be pursued for all of this, but this one does the trick for now.

As for tests: I think we'd need more testing apparatus to really test this one (have to extract output from ``setup.py`` commands, which it looks like ``run_setup`` doesn't currently do). I think a test would be good, but I wanted to put the fix out here first, since this is a really confusing bug.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,9 @@ astropy-helpers Changelog
   convenience. Version shipped with astropy-helpers is v0.5.
   [#278, #303, #309]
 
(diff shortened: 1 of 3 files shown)


bsipocz:
Rebased and adding changelog. While I agree that it would be great to test this, having the fix in is already an improvement.

PR #295, opened 2017-04-11
Python intersphinx target has moved to https

MSeifert04:
This should fix the documentation "warning" I've seen in [`ccdproc`](https://travis-ci.org/astropy/ccdproc/jobs/221047731#L557):

```
loading intersphinx inventory from http://docs.python.org/3/objects.inv...
intersphinx inventory has moved: http://docs.python.org/3/objects.inv -> https://docs.python.org/3/objects.inv
```


bsipocz:
Thanks @MSeifert04!

Issue #305, opened 2017-05-29
pyflakes issue pointed out in astropy_helpers/distutils_helpers.py


cdeil:
@bsipocz - This is a real bug, no?
```
$ python -m pyflakes astropy_helpers/ | egrep -v star
astropy_helpers/distutils_helpers.py:193: undefined name 'log'
```

Issue #302, opened 2017-05-19
Issues after update to Sphinx 1.6.1


cdeil:
The Gammapy docs build broke with the update to the new Sphinx 1.6.1.
I didn't check yet for Astropy or other affiliated packages.

First I thought it's a regression in Sphinx, so I opened https://github.com/sphinx-doc/sphinx/issues/3766 yesterday. But I'm not sure if this is really a bug / regression in Sphinx, or if it's something in `astropy-helpers` that now surfaced because it relies on Sphinx internals (the traceback there mentions `AstropyBuildSphinx`).

Locally when I update to Sphinx 1.6.1 I see a different error:
```
Traceback (most recent call last):
  File "<stdin>", line 36, in <module>
  File "/Users/deil/Library/Python/3.6/lib/python/site-packages/sphinx/application.py", line 233, in __init__
    self._init_builder()
  File "/Users/deil/Library/Python/3.6/lib/python/site-packages/sphinx/application.py", line 311, in _init_builder
    self.emit('builder-inited')
  File "/Users/deil/Library/Python/3.6/lib/python/site-packages/sphinx/application.py", line 488, in emit
    return self.events.emit(event, self, *args)
  File "/Users/deil/Library/Python/3.6/lib/python/site-packages/sphinx/events.py", line 79, in emit
    results.append(callback(*args))
  File "/Users/deil/code/gammapy/astropy_helpers/astropy_helpers/sphinx/ext/automodsumm.py", line 252, in process_automodsumm_generation
    lines = automodsumm_to_autosummary_lines(sfn, app)
  File "/Users/deil/code/gammapy/astropy_helpers/astropy_helpers/sphinx/ext/automodsumm.py", line 311, in automodsumm_to_autosummary_lines
    if 'astropy_helpers.sphinx.ext.automodapi' in app._extensions:
AttributeError: 'Sphinx' object has no attribute '_extensions'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 42, in <module>
NameError: name 'self' is not defined
Sphinx Documentation subprocess failed with return code 1
```

Looks like `astropy-helpers` accesses a private `_extensions` property in Sphinx and that now broke with Sphinx 1.6.1?

cdeil:
I see the same error in the Astropy master Sphinx build, i.e. I think any dev or CI updating to Sphinx 1.6.1 will run into this issue.

saimn:
[...]

PR #306, opened 2017-05-30
Fix errors reported by pyflakes

astrofrog:
Fixes https://github.com/astropy/astropy-helpers/issues/305 - I also found another error that I fixed here, and fixed some other more minor issues.

### DIFF
--- a/astropy_helpers/commands/_test_compat.py
+++ b/astropy_helpers/commands/_test_compat.py
@@ -145,7 +145,7 @@ class AstropyTest(Command, object):
         running the tests.
         """
         try:
-            import astropy
+            import astropy  # noqa
         except ImportError:
             raise ImportError(
                 "The 'test' command requires the astropy package to be "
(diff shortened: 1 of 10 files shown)


bsipocz:
Thanks @astrofrog. It all looks good to me.

Do you think it should go into 1.3.2, or 2.0? My best guess is that the current 1.3.x branch has issues with setuptools when a too new version is picked up from conda-forge (at least I see the jinja2 issues only on travis for both gammapy/photutils/sunpy, some are linked in #302), but not locally where I have an ancient version. I'm now testing this hypothesis on travis.

astrofrog:
I would say 1.3.2 since it fixes actual potential bugs

bsipocz:
Since this is mostly for affiliates that doesn't want to upgrade to v2.0 straight away, I've released astropy core 1.3.3 without the docs there needs some other fixes, too to be sphinx 1.6 compatible.

The 1.3.x branch here is currently failing with gammapy/photutils but passed with astropy. So anything can still be backported.

bsipocz:
Why wasn't astropy-bot complaining about the missing milestone and changelog?

PR #308, opened 2017-05-30
Fix compatibility with latest version of setuptools

astrofrog:
This gets the build running but there are real errors, so I will investigate further.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,26 +23,15 @@ env:
     - PYTHON_VERSION=3.4
     - PYTHON_VERSION=3.5
     - PYTHON_VERSION=3.6 EVENT_TYPE='push pull_request cron'
-
+    - PYTHON_VERSION=2.7 SETUPTOOLS_VERSION=dev DEBUG=True
+      CONDA_DEPENDENCIES='pytest sphinx cython numpy packaging six appdirs'
+      EVENT_TYPE='push pull_request cron'
+    - PYTHON_VERSION=2.7 SETUPTOOLS_VERSION=20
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -23,29 +23,15 @@ env:
     - PYTHON_VERSION=3.4
     - PYTHON_VERSION=3.5
     - PYTHON_VERSION=3.6 EVENT_TYPE='push pull_request cron'
-
+    - PYTHON_VERSION=2.7 SETUPTOOLS_VERSION=dev DEBUG=True
+      CONDA_DEPENDENCIES='pytest sphinx cython numpy packaging six appdirs'
```
bsipocz: pytest and pip are default installs by ci-helpers, no need to list them again

bsipocz:
@astrofrog - Is this ready to be merged?

.travis.yml
astrofrog: Fixed

astrofrog:
@bsipocz - yes, once tests pass (had to rebase)

bsipocz:
I'll try to backport this to 1.3.2, though doubt that it would solve any of the issues...

bsipocz:
@astrofrog - Is it necessary to test everything on both osx and linux?

astrofrog:
@bsipocz - I've seen some setuptools issues with just MacOS X, so I was just trying to be on the safe side, but we could potentially reduce the number of builds per platforms (e.g. the SETUPTOOLS_VERSION=20 build could maybe be combined with one of the other ones)

bsipocz:
OK, that sounds good. I'll also move the sphinx 1.6+ build to one platform as that one is only a temporary one.

PR #309, opened 2017-05-30
Updating bundled sphinx-automodapi to v0.5

bsipocz:


bsipocz:
Once this is merged, I'd like to open a PR to astropy to see how it behaves with the dev branch here. Yesterday I saw a few new docs errors, so it would be better to iron out any issues well before feature freeze.

PR #314, opened 2017-06-04
Remove deprecated _test_compat

### LINKED ISSUE #307: Remove _test_compat.py?
The old implementation of the ``python setup.py test`` command is kept in ``astropy_helpers.commands._test_compat``, and this has been deprecated as of Astropy/astropy-helpers 1.1. We might want to consider cleaning that up soon?

bsipocz:
To fix #307


bsipocz:
Hmm, I guess it would be useful to update the error message, too to mention the minimal astropy version rather than just saying "error: The 'test' command requires the astropy package to be installed and importable."

bsipocz:
@eteq @astrofrog - While I was doing the RC testing, it became apparent that with this removal, the helpers is not compatible with astropy <1.1 any more, and requires 1.1+ to be installed. Are we happy with that for 2.0, or shall postpone it to 3.0?

astrofrog:
I'm trying to think in what cases that might matter in practice - what are your thoughts?

bsipocz:
Well, it means that all packages should drop supporting astropy 1.0.x when switching to this helper in order to make their testing work. I'm not sure how many packages plan to keep supporting 1.0 for much longer, they then definitely cannot update to the 2.0 helpers. E.g. ccdproc still seems to have plans to support 1.0.x for a while. Maybe @mwcraig can give his opinion here.

It may be also possible it can be an issue for a package that has astropy only as an optional dependency (glue is one of these, but you don't use the helpers there). I though that ginga is also in this category but nevertherless it also lists astropy in its ``install_requires``.

mwcraig:
@bsipocz -- if we don't test against 1.0.x anymore will the remaining tests pass? We don't intend to support astropy 1.0.x for very long once 2.0 is out....roughly a month or two, and we'd be happy not to test against it any more.

mwcraig:
So I think what we are holding off on is a change in the minimum required version of astropy.

bsipocz:
@mwcraig - I've tried it in photutils, and everything except LTS is passed. Can try in ccdproc, too.

1) So the two options, ship the RC as is meaning that helpers 2.0 will require astropy 1.1+

2) don't include this change in helpers 2.0 but postpone to 3.0. That time we probably get the testing stuff moved out of astropy to a separate package (https://github.com/astropy/astropy/pull/5770) meaning that the helpers won't depend on astropy, but on the separate, much less complex testing package.

While I like removing deprecated code sooner rather than later, I slightly inclined to option 2) to avoid possible issues and complaints from packages that are not dependent on astropy (even though I didn't find much of these yet).

mwcraig:
I agree that 2 sounds like the better option, that way there is plenty of time for the affiliated packages to update.

astrofrog:
I'm in favor of option 2

astrofrog:
I don't think https://github.com/astropy/astropy/pull/5770 is about moving the testing stuff into a separate package though, just putting each plugin in a separate sub-module in astropy?

bsipocz:
In my understanding astropy/astropy#5770, or the bigger picture is about moving it out of astropy core, to a pytest plugin package. At least that was the gsoc plan, with the motivation from @Cadair that the testing framework has nothing to do with astropy thus it shouldn't be a dependency for packages that would only need the testing infrastructure.

astrofrog:
Ah right I see (read the issue in more detail)

pllim:
c/c @ejeschke for his information on future Ginga support.

bsipocz:
This is now removed from the 2.0.x branch, so we'll go with option 2) meaning a seamless helpers update for the affiliated packages.

Issue #297, opened 2017-04-12
ez_setup is deprecated


bsipocz:
Current version of ez_setup.py comes with a big deprecation at the top, so we should also revise its usage here ideally by the 2.0 release.

astrofrog:
Have they suggested a replacement?

bsipocz:
It seems that pip installing, but I haven't looked into the very details yet. Our build against the dev version is hopelessly broken though, but it may be an issue upstream as I couldn't hack it last night.

The last ez_setup version is for setuptools 33.1.1, which is fairly recent, so as a workaround we could stick with that for a while (conda still ships v27.x).

saimn:
Interesting, so from I can read:
- They dropped self-upgrade support in setuptools (https://github.com/pypa/setuptools/issues/581)
- With [PEP-518](https://www.python.org/dev/peps/pep-0518/) it is (will be ?) possible to tell pip the required build dependencies, with a `pyproject.toml` file. 
- So I guess the idea to replace `ez_setup` is simply to tell pip that (a recent) setuptools is required. But I don't see what should be done for users with an old pip, other than telling them to upgrade pip ?
- An interesting related point is that it is now possible to configure `setup()` using `setup.cfg` files, in a way similar to what astropy_helpers/package_template do (or pbr, d2to1): https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files . If it become easy to require a recent version of setuptools, this looks great.

embray:
I noticed that last bullet point recently as well and thought of bringing it up here.  The problem is, as you say, it requires a recent version of setuptools.  What one might be able to do is move that functionality to astropy-helpers (if it hasn't been already) and just enable it when needed for backwards-compatibility.  Then most package maintainers will almost never have to touch setup.py itself which is a good thing.

bsipocz:
For 2.0.0 we plan to ship with latest ``ez_setup.py``, but will get rid of it in 3.0.0

astrofrog:
I think the path forward for ``ez_setup.py`` is that we should remove it, and also update the astropy installation docs so that users installing locally should use ``pip install .`` rather than ``python setup.py install``. Recent versions of pip should install setuptools automatically, but for older versions I think most users will anyway have setuptools installed in their environments, so I think this isn't a big deal.

astrofrog:
[...]

PR #321, opened 2017-06-22
Updating ez_setup to the last version of 33.1.1

bsipocz:
It is deprecated and removed in later versions, but we plan to address that issue in 3.0.0

I'm not sure whether the deprecation warning will cause any troubles along the line, if it has issues during RC testing we should consider reverting this PR in the actual release.


bsipocz:
For discussion about the phasing out of ez_setup, see #297

Issue #320, opened 2017-06-22
Try to loosen the install-time sphinx dependencies


eteq:
@josePhoenix  reported a problem in #319 that can be traced back to sphinx doing an update that broke some of the astropy sphinx machinery in a way that made even a non-doc build fail.  While that *specific* issue has been resolved in the current helpers, it begs a question: can we "protect" the sphinx imports better in some way so that if they fail the exceptions don't kill the whole build process?

Right now if sphinx isn't installed this is caught, but a *failing* sphinx apparently is not.  So perhaps we can  add some `try`/`except` blocks in clever places to convert exceptions into warnings *if* we're not in a ``build_docs`` comamnd?  I'm not sure exactly how much work that will be, but I'm raising it now in the hopes that I or someone else can get to it later.

pllim:
For the particular problem that raised this issue, Sphinx was installed (if it wasn't, the error would not have occurred) but was the incompatible 1.6 version *and* `astropy-helpers` submodule was already fixed by whatever Astropy that was also installed. There is nothing we can do there.

As for future-proofing, we can never predict what is going to change in Sphinx... In this case, the usual `try ... except ...` would be:
```python
try:
    from sphinx.setup_command import BuildDoc as SphinxBuildDoc
except ImportError:
    do_magic
```
but that would not have caught `RecursionError` (and who would think of catching that in an import statement?).

This is tricky. :thinking:

josePhoenix:
I use other packages that depend on Sphinx, and none of those builds broke. So, the question is, what are they doing differently?

A related question would be, how do the astropy helpers cause RecursionErrors in 3rd party packages?

pllim:
`astropy-helpers` add `build_docs` setup command to `setup.py`, so you don't see the error in other packages because they are not as fancy? (Just speculating since I don't know what other packages you are referring to.)

"How" and "why" are good questions... They are why we still have jobs. :sweat_smile:

saimn:
There is already a try/except that skip registering the `build_docs` command if Sphinx is not installed:
https://github.com/astropy/astropy-helpers/blob/master/astropy_helpers/setup_helpers.py#L141

But the recent issue was triggered with Sphinx installed but the import failing in a non-obvious way. It would be easy to catch any exception, and it would avoid breaking the setup.py for users, but it would also hide a bit the issue for developers.

bsipocz:
[...]

PR #322, opened 2017-06-22
Adding sphinx dev testing to the travis matrix

bsipocz:
As discussed in #320, we should consider test against the sphinx dev to spot issues early on.

### DIFF
--- a/astropy_helpers/sphinx/ext/doctest.py
+++ b/astropy_helpers/sphinx/ext/doctest.py
@@ -11,7 +11,7 @@ tests.
 """
 import re
 from docutils.nodes import literal_block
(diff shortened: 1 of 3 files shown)


astropy_helpers/sphinx/ext/doctest.py
```diff
@@ -11,7 +11,7 @@
 """
 import re
 from docutils.nodes import literal_block
-from sphinx.util.compat import Directive
+from docutils.parsers.rst import Directive
```
bsipocz: I think changing this line is safe as it's around at least since 2012 and now being removed from sphinx.

Issue #325, opened 2017-07-10
2.0: Does not work with sphinx 1.4.9?


olebole:
Hi,

after uploading 2.0, I got a bug report [debian/#867772](https://bugs.debian.org/867772) that python-drizzle (from STSCI) [could not be built](https://tests.reproducible-builds.org/debian/rbuild/unstable/amd64/python-drizzle_1.5-2.rbuild.log).
```
Traceback (most recent call last):
  File "setup.py", line 123, in <module>
    **package_info
  File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/usr/lib/python2.7/dist-packages/astropy_helpers/commands/build_sphinx.py", line 260, in run
    AstropyBuildDocs.run(self)
  File "/usr/lib/python2.7/dist-packages/astropy_helpers/commands/build_sphinx.py", line 170, in run
    """).format(pdb=self.pdb, verbosity=self.verbosity,
  File "/usr/lib/python2.7/distutils/cmd.py", line 105, in __getattr__
    raise AttributeError, attr
AttributeError: pdb
```
That was with sphinx 1.4.9. Now, when Debian has upgraded to sphinx 1.5.6, the problem disappeared.
Just for completeness (and possible backporting): Is sphinx-1.4 support silently dropped?

bsipocz:
It may be a side effect of one of the PRs, let me dig into it.

bsipocz:
this is closed by #326

PR #326, opened 2017-07-10
Fixing compatibility with sphinx <1.5

bsipocz:
#292 seems to broke our compatibility with older sphinx versions. We simply need to test against 1.3, 1.4 and 1.5 to make sure or update the current requirement of ``1.3``

See: #325


bsipocz:
@astrofrog - the bot is malfunctioning, there is a changelog here...

astrofrog:
I will investigate!

pllim:
Maybe replacing existing line doesn't count as "new change log entry" by the bot's rules. 🤔

bsipocz:
@pllim - that shouldn't matter I guess. Also previous reincarnation of the bot worked (https://github.com/astropy/astropy-helpers/pull/323#issuecomment-313047183), so probably one of my PRs messed it up.

pllim:
My bet is on https://github.com/astropy/astropy-bot/blob/master/changebot/changelog.py#L33 because I notice that this change log here has "====" at the top but that line does not contain version info like Astropy's. Just a guess though.

pllim:
Bot is happy now! 😄

bsipocz:
Thanks @pllim!

astrofrog:
I'm working on an updated version of the bot that won't require this kind of hack

bsipocz:
@astrofrog @eteq - What do you think how critical breaking sphinx compatibility is (1.4 is not that old, the latest of that series was release at the end of last year)? Should we aim to have a helpers 2.0.1 along with astropy 2.0.1?

astrofrog:
I think it makes sense to keep compatibility with at least 1.4+

eteq:
I also agree compatibility with something that recent is fairly important, so it's worth a release.

Issue #317, opened 2017-06-19
Warnings during CI testing


bsipocz:
New pytest 3.1+ prints out warnings gathered during running the test session. In the long term it would be nice to fix these when possible.

```
=============================== warnings summary ===============================

astropy_helpers/tests/test_git_helpers.py::test_installed_git_version[1.0.dev]

  /Users/travis/miniconda/envs/test/lib/python2.7/site-packages/setuptools-27.2.0-py2.7.egg/setuptools/dist.py:331: UserWarning: Normalizing '1.0.dev' to '1.0.dev0'

astropy_helpers/tests/test_setup_helpers.py::test_compiler_module

  apyhtest_eva/setup_package.py:1: RuntimeWarning: Parent module 'apyhtest_eva' not found while handling absolute import

  apyhtest_eva/setup_package.py:2: RuntimeWarning: Parent module 'apyhtest_eva' not found while handling absolute import

astropy_helpers/tests/test_setup_helpers.py::test_no_cython_buildext

  apyhtest_eva/setup_package.py:1: RuntimeWarning: Parent module 'apyhtest_eva' not found while handling absolute import

  apyhtest_eva/setup_package.py:2: RuntimeWarning: Parent module 'apyhtest_eva' not found while handling absolute import

astropy_helpers/tests/test_setup_helpers.py::test_missing_cython_c_files

  apyhtest_eva/setup_package.py:1: RuntimeWarning: Parent module 'apyhtest_eva' not found while handling absolute import

  apyhtest_eva/setup_package.py:2: RuntimeWarning: Parent module 'apyhtest_eva' not found while handling absolute import

astropy_helpers/tests/test_setup_helpers.py::test_build_docs[deprecated]

  /Users/travis/build/bsipocz/astropy-helpers/astropy_helpers/commands/build_sphinx.py:259: AstropyDeprecationWarning: The "build_sphinx" command is now deprecated. Use"build_docs" instead.

    '"build_docs" instead.', AstropyDeprecationWarning)
```

astrofrog:
I don't see anymore warnings in the CI since #330, so closing

PR #330, opened 2017-08-15
Fix warnings

drdavella:
This partially addresses #317. In particular, these two warnings no longer occur:

>astropy_helpers/tests/test_git_helpers.py::test_installed_git_version[1.0.dev]
>
>  /Users/travis/miniconda/envs/test/lib/python2.7/site-packages/setuptools-27.2.0-py2.7.egg/setuptools/dist.py:331: UserWarning: Normalizing '1.0.dev' to '1.0.dev0'
>
> astropy_helpers/tests/test_setup_helpers.py::test_build_docs[deprecated]
>
> /Users/travis/build/bsipocz/astropy-helpers/astropy_helpers/commands/build_sphinx.py:259: AstropyDeprecationWarning: The "build_sphinx" command is now deprecated. Use"build_docs" instead.
>
>    '"build_docs" instead.', AstropyDeprecationWarning)

I wasn't able to reproduce the other warnings on my own system, so I'm not sure whether they only occur in particular environments or whether they have already been resolved.

### DIFF
--- a/astropy_helpers/tests/test_git_helpers.py
+++ b/astropy_helpers/tests/test_git_helpers.py
@@ -241,9 +241,10 @@ def test_installed_git_version(version_test_package, version, tmpdir, capsys):
 
 
 def test_get_git_devstr(tmpdir):
+    dirpath = str(tmpdir)
     with catch_warnings(record=True) as w:
-        devstr = get_git_devstr(path=tmpdir)
+        devstr = get_git_devstr(path=dirpath)
         assert devstr == '0'
         assert len(w) == 1
         assert str(w[0].message).startswith(
-            "No git repository present at local('{}')".format(tmpdir))
+            "No git repository present at '{}'".format(dirpath))
--- a/astropy_helpers/tests/test_git_helpers.py
+++ b/astropy_helpers/tests/test_git_helpers.py
@@ -242,9 +242,13 @@ def test_installed_git_version(version_test_package, version, tmpdir, capsys):
 
 def test_get_git_devstr(tmpdir):
     dirpath = str(tmpdir)
+    warn_msg = "No git repository present at"
+    # Verify as much as possible, but avoid dealing with paths on windows
+    if not sys.platform.startswith('win'):
+        warn_msg += " '{}'".format(dirpath)
+
     with catch_warnings(record=True) as w:
         devstr = get_git_devstr(path=dirpath)
         assert devstr == '0'
         assert len(w) == 1
-        assert str(w[0].message).startswith(
-            "No git repository present at '{}'".format(dirpath))
+        assert str(w[0].message).startswith(warn_msg)
--- a/astropy_helpers/tests/test_git_helpers.py
+++ b/astropy_helpers/tests/test_git_helpers.py
@@ -7,9 +7,11 @@ import sys
 import tarfile
 
 import pytest
+from warnings import catch_warnings
 
 from . import reset_setup_helpers, reset_distutils_log, fix_hide_setuptools  # noqa
 from . import run_cmd, run_setup, cleanup_import
+from astropy_helpers.git_helpers import get_git_devstr
 
 
 PY3 = sys.version_info[0] == 3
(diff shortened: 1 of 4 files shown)


bsipocz:
@astrofrog - The bot got pretty slow at commenting. Is that expected? Can you see any issues in the logs?

astropy_helpers/tests/test_git_helpers.py
```diff
@@ -37,6 +37,12 @@
 
 if not RELEASE:
     VERSION += get_git_devstr(False)
+    # This workaround is required to account for the fact that we test
+    # installation of this package in test_installed_git_version even with
+    # development versions. In these cases get_git_devstr does not work since
+    # it is called from outside of a git repository.
+    if VERSION.endswith('.dev'):
+        VERSION = VERSION.replace('.dev', '')
```
astrofrog: If ``get_git_devstr`` doesn't work, shouldn't the version number retain at least the ``.dev`` suffix so it doesn't look like a stable version?

drdavella: I'm not sure it matters much since all of this is just within a test sandbox, but another possible solution would be to do something like this instead:
```python
if VERSION.endswith('.dev'):
    VERSION += '0'
```

pllim: It does matter for developers. I have multiple environments with different dev versions. So, I would like that info retained as well.

drdavella: @pllim, agreed, but the only reason there is a problem here is because the test itself is running in a sandbox outside of a git tree, and this fix is solely intended to address that problem. This isn't an issue that a developer should ever experience directly, unless I'm missing something.

pllim: I see it even when running from a `git` source checkout for some of my projects but I never bothered to investigate. Somehow the final `pytest` report still gave me the correct dev string. Anyway, if this version string is going into a pytest report, it is less confusing to have a "dev" still retained. Otherwise, people might think the test result is from a released version. Or am I misunderstanding something here?

drdavella: This could just as easily have been addressed in the call to `pytest.mark.parametrize` on line 190, by changing `'1.0.dev'` to `'1.0.dev0'` and getting rid of the call to `get_git_devstr` altogether. That might be cleaner, and maybe I should do it.

The problem with that approach is I guess we never actually test `get_git_devstr` in that case, but it's really not possible to test its proper behavior if the tests always run in a temp directory outside of the source tree.

astrofrog: @drdavella - if someone downloads the dev version of astropy from git as a zip file they will end up (I think) with the full dev version but no ``.git`` directory.

drdavella: The dev string in this test will not be reported by pytest. This is part of a meta-test of installing a fake package and making sure astropy_helpers tags the installed version with the correct git hash.

When you and I run setup.py directly, it's always within a git tree. The setup.py under question here is generated as part of a fake package that is used for testing purposes. At one point in the test, a tarred up version of the package is extracted into a temporary directory. There's no longer a git tree here, which is why the 'dev' VERSION string doesn't get a revision number. I believe this was just an oversight from when the test was first designed, and my commit just tries to correct for it.

pllim: Ohhh... Just the *test*... Sorry for the noise. I obviously need more coffee. Will promptly remedy that right now. ☕️

drdavella: @astrofrog, I guess that's a use case I hadn't considered. In that case, I'm not sure what the right solution should be. It's possible `get_git_devstr` could just return '0' in the absence of a git repository. That would avoid the warning from `setuptools`, but maybe in that case `get_git_devstr` should issue its own warning indicating that it's running outside of a git repo.

astrofrog: > It's possible get_git_devstr could just return '0' in the absence of a git repository. That would avoid the warning from setuptools, but maybe in that case get_git_devstr should issue its own warning indicating that it's running outside of a git repo.

That sounds like the right approach

drdavella:
This probably needs a (different) change log entry since it now includes a change to user-facing behavior.

astrofrog:
@drdavella - agree, but I think you should mention the change in behavior of the git dev version

astropy_helpers/tests/test_setup_helpers.py
```diff
@@ -1,7 +1,9 @@
 import os
 import sys
+import six
```
bsipocz: Minor point, but as I recall we didn't required six before, so maybe a hack for the version number below would be slightly better.

bsipocz:
Looks good to me except the nitpicky comment that may be overcautious anyway.

Issue #329, opened 2017-08-09
Hidden package caching behavior breaks package exclusion


drdavella:
The `get_package_info` command provides an optional parameter `exclude=`, which tells the `find_packages` routine in `setuptools` to ignore packages with names that match certain patterns. For example, a common use case would be to avoid installing tests:

```python
# setup.py
import ah_bootstrap
from setuptools import setup
from astropy_helpers.setup_helpers import get_package_info

...

package_info = get_package_info(exclude=['*.tests'])
setup(**package_info)
```

In many (if not all) cases, a project will also want to use custom commands provided by `astropy_helpers`. To do this, the code ends up looking like the following:

```python
# setup.py
import ah_bootstrap
from setuptools import setup
from astropy_helpers.setup_helpers import get_package_info, register_commands

...

release = False
cmdclassd = register_commands('foobar', '1.0.dev', release)
package_info = get_package_info(exclude=['*.tests'])
setup(cmdclass=cmdclassd, **package_info)
```

Here's the problem: when I call `register_commands`, it implicitly calls the underlying `find_packages` routine, but without any knowledge of the packages that I want to exclude. It then does me the favor of caching the packages that it found so that it doesn't have to search again. However, when I call `get_package_info`, it goes straight to the cached list of packages and does not filter against my requested exclusions.

I think I understand the issue well enough so I should be able to submit a PR with a fix sooner or later. Any input on the right approach is welcome, though. Right now it's not clear to me whether `register_commands` actually needs to know anything about the listed packages or not. If it doesn't, that would be an easy and clean fix.

pllim:
Pinging @embray just in case he can comment on this... 🙏

drdavella:
A possible solution could be to add an `exclude` parameter to `register_commands`, but it's not exactly intuitive and seems prone to user error. Another solution would be to disable caching when `find_packages` is called by `register_commands`, but that seems to defeat the purpose. It's not a hard fix, it's just a matter of finding the _right_ fix. :)

drdavella:
This was closed by #331.

PR #331, opened 2017-08-15
Fix broken package exclusion caching behavior

drdavella:
Taking a crack at fixing #329. Feedback is welcome on whether this is a reasonable fix or whether we should use another approach. If this seems reasonable, it's probably a good idea to add another test to make sure the package exclusion mechanism actually works.

This is a bit touchy because it involves user-facing APIs that don't actually work as expected. I added deprecation warnings for the affected parameters but they really shouldn't be used at all because they don't work. Maybe someone else has a better idea of how to handle this.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,10 @@ astropy-helpers Changelog
 
 - Switch to using mathjax instead of imgmath for local builds. [#342]
 
+- Deprecate ``exclude`` parameter of various functions in setup_helpers since
+  it could not work as intended. Add new function ``add_exclude_packages`` to
+  provide intended behavior. [#331]
+
 2.0.1 (2017-07-28)
 ------------------
 
--- a/astropy_helpers/tests/test_setup_helpers.py
+++ b/astropy_helpers/tests/test_setup_helpers.py
@@ -543,3 +543,65 @@ def test_adjust_compiler(monkeypatch, tmpdir):
 
         assert len(log.messages) == 1
         assert 'The C compiler used to compile Python' in log.messages[0]
+
+
+def test_invalid_package_exclusion(tmpdir, capsys):
+
+    module_name = 'foobar'
+    setup_header = dedent("""\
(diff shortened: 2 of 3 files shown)


bsipocz:
@eteq - can we milestone this to 2.0.2, or 3.0 is more appropriate? Strictly speaking this is an API change, but on the other hand it fixes a bug.

drdavella:
@bsipocz just rebased this one since it was a bit stale. Any ideas about milestones or whether this can be merged soon?

astropy_helpers/tests/test_setup_helpers.py
```diff
@@ -543,3 +543,196 @@ def compiler_setter_with_sysconfig(compiler):
 
         assert len(log.messages) == 1
         assert 'The C compiler used to compile Python' in log.messages[0]
+
+
+def test_adjust_compiler(monkeypatch, tmpdir):
```
astrofrog: Why is this test repeated?

CHANGES.rst
```diff
@@ -30,6 +30,9 @@ astropy-helpers Changelog
 
 - Switch to using mathjax instead of imgmath for local builds. [#342]
 
+- Fix incorrect behavior when using ``exclude`` parameter of various functions
+  in setup_helpers. [#331]
```
astrofrog: The code change suggest exclude is actually deprecated rather than fixed?

bsipocz:
I've added the 2.0.2 milestone for now, @astrofrog may change it if he thinks otherwise.

astropy_helpers/tests/test_setup_helpers.py
drdavella: Good catch, I think it's a copy/paste error.

CHANGES.rst
drdavella: You're right, I've clarified it in an updated entry.

bsipocz:
Thanks @drdavella!

PR #335, opened 2017-08-31
Allow Sphinx doctest extension to recognize standard doctest directives

drdavella:
I started looking at https://github.com/astropy/astropy/issues/6010 and realized that it would be quite useful to be able to use the doctest directives `testsetup` and `doctest`. However, when building the documentation, Sphinx doesn't recognize these directives since we're not actually using  the Sphinx doctest plugin, but a simple custom extension that we provide ourselves.

This is a fairly simple change that allows the `testsetup` and `doctest` directives to be recognized and handled properly when building Sphinx documentation. It's not exhaustive since I don't have a need for the other cases at the moment. However, it doesn't seem difficult to add support for [the others](http://www.sphinx-doc.org/en/stable/ext/doctest.html#directives) as well.

### DIFF
--- a/astropy_helpers/sphinx/ext/doctest.py
+++ b/astropy_helpers/sphinx/ext/doctest.py
@@ -37,6 +37,7 @@ def setup(app):
     app.add_directive('doctest-requires', DoctestRequiresDirective)
     app.add_directive('doctest-skip', DoctestSkipDirective)
     app.add_directive('doctest-skip-all', DoctestSkipDirective)
+    app.add_directive('doctest', DoctestSkipDirective)
 
     return {'parallel_read_safe': True,
             'parallel_write_safe': True}
--- a/astropy_helpers/sphinx/ext/doctest.py
+++ b/astropy_helpers/sphinx/ext/doctest.py
@@ -26,6 +26,14 @@ class DoctestSkipDirective(Directive):
         return [literal_block(code, code)]
 
 
+class DoctestOmitDirective(Directive):
+    has_content = True
+
+    def run(self):
+        # Simply do not add any content when this directive is encountered
+        return []
+
+
 class DoctestRequiresDirective(DoctestSkipDirective):
     # This is silly, but we really support an unbounded number of
     # optional arguments
@@ -38,6 +46,7 @@ def setup(app):
     app.add_directive('doctest-skip', DoctestSkipDirective)
     app.add_directive('doctest-skip-all', DoctestSkipDirective)
     app.add_directive('doctest', DoctestSkipDirective)
+    app.add_directive('testsetup', DoctestOmitDirective)
 
     return {'parallel_read_safe': True,
             'parallel_write_safe': True}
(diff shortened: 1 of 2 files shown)


pllim:
Astropy's custom doctest extension supports things like "skip remote data" or "relaxed floating point comparison". From a glance, I don't see these supported here, which would cause doctest to fail. Will you add more to this PR?

drdavella:
@pllim, those are inline directives, right? This PR affects only block-level directives like the following:
```rst
Hey look at this example!

.. testsetup::
>>> # This code won't actually appear in the documentation but makes the example possible
>>> thing = dict(a=0,b=1,c=2)

.. doctest::
>>> # This code does appear in the documentation! Even this comment!
>>> print(thing['a'])
```

The problem is that Sphinx doesn't recognize these block-level directives since we don't actually use Sphinx's built-in doctest plugin. However, when we run our own doctest plugin through pytest, any inline directives like the ones you mentioned should be handled whether or not they are in one of these new blocks or not.

Maybe I've misunderstood or am missing something, though, so let me know if that doesn't address your question.

pllim:
I discussed with @drdavella offline and agreed that `testsetup` might be useful when a dataset needs to be *regenerated* on-the-fly for narrative continuation in a different page. However, it is not a good idea to completely hide how an input is first generated, as the doc narrative needs to be reproducible.

bsipocz:
@astrofrog - would you think this is 2.0.2 or 3.0?

drdavella:
It's worth mentioning that an equally valid (and probably cleaner) solution to this problem would be to just make use of the Sphinx doctest extension so that these directives are recognized and handled. 

Apparently it's possible to set `doctest_test_doctest_blocks = None` in `conf.py` so that Sphinx still won't actually run the doctests.

bsipocz:
Could you @drdavella rebase?

drdavella:
@bsipocz done

astropy_helpers/sphinx/ext/doctest.py
```diff
@@ -37,6 +45,8 @@ def setup(app):
     app.add_directive('doctest-requires', DoctestRequiresDirective)
     app.add_directive('doctest-skip', DoctestSkipDirective)
     app.add_directive('doctest-skip-all', DoctestSkipDirective)
+    app.add_directive('doctest', DoctestSkipDirective)
+    app.add_directive('testsetup', DoctestOmitDirective)
```
astrofrog: Could you add a comment here explaining how ``DoctestSkipDirective`` is different from ``DoctestOmitDirective``?

drdavella:
@astrofrog the motivation for this is below, but **TL;DR** there's no longer a pressing use case for this and so it might be best just to close the PR.

These are directives that are actually provided by the sphinx doctest plugin, but since we don't use that, they aren't recognized by sphinx when we build the documentation. The idea is that you can hide certain boilerplate code that is only used for the purposes of testing your documentation, but does not need to be included in the generated documentation itself. 

The intent here is not to enable sphinx itself to test these blocks of code. To do that, we would have to install sphinx's doctest plugin, in which case these directives would already be defined. Instead, the point is to allow docs with these directives to be understood by sphinx during the build phase, so that we can use the directives if we want when we run doctest with pytest.

Originally I thought this would be useful for the FITS documentation, but it's not being used currently so there's not a strong motivation to merge. However, I think it doesn't doesn't hurt anything, so it could still be merged if there are no objections.

bsipocz:
arrg, could you rebase with a `[skip ci]`? there is a conflict with the changelog...

bsipocz:
Thanks @drdavella!

PR #340, opened 2017-09-26
Removing python2 support

### LINKED ISSUE #333: Deprecate or remove requires_2to3 support
This is probably a discussion for the in-person coordination meeting, but I think we should consider removing support for ``requires_2to3`` (which allows 2to3 to be disabled on a per-subpackage basis) once we make sure that no affiliated packages are using it. If we make this decision, then I think the following things can be removed:

* The custom ``build`` command (``AstropyBuildPy``), because the main purpose of this command is to deal with the 2to3 option and also to separate build directories for different Python versions even for pure-Python packages, which was only needed because the Python 2 and 3 source code could be different.

* The custom ``install`` and ``install_lib`` commands (``AstropyInstall`` and ``AstropyInstallLib``) - these were only needed because we were customizing the build location with ``AstropyBuildPy``, but should no longer be needed once we use standard build directories.

* Some code in ``setup_helpers.py`` related to this option.

Overall, the main benefit would be to go back to not providing custom build, install, and install_lib commands. In general I think we should always be striving to use the standard mechanisms for package maintenance whenever possible.

Maybe we could set a deadline of 3.0 for this, and warn affiliated package maintainers so that they adjust their code base to be 2 and 3-compatible? (which would be a good thing anyway). Note that if they don't want to do that yet, they can simply avoid updating to a more recent astropy-helpers.

### LINKED ISSUE #337: Remove python2 support
In coordination with astropy core 3.0, we can/should remove any python2 related code from here.

bsipocz:
This is to close #337 and close #333

TODO (please add to the list any more items, I've opened this as a WIP to see CI):
- [x] Add detailed warning in the README, that affiliates shouldn't update to helpers 3.0 if they wish to keep pthon2 support
- [x] need to be tested against astropy (and if possible against at least one affiliate package) before merge.


bsipocz:
@astrofrog @pllim - Something is not right with the bot, it doesn't comment here any more.

astropy_helpers/commands/build_py.py
```diff
@@ -21,19 +21,6 @@ def finalize_options(self):
...
-        filtered_files = []
-        for filename in files:
-            for package in skip_2to3:
-                if filename[len(self.build_lib) + 1:].startswith(package):
-                    break
-            else:
-                filtered_files.append(filename)
-
-        SetuptoolsBuildPy.run_2to3(self, filtered_files, doctests)
-
```
astrofrog: I think the whole file can be removed - the splitting of the build directories was done because of Python 2/3 but that actually causes problems elsewhere. Since this isn't needed anymore, I suggest removing the whole custom build command.

astrofrog:
The custom build, install, and install_lib commands can be removed completely I think

astropy_helpers/commands/_dummy.py
```diff
@@ -69,13 +69,7 @@ def __getattribute__(cls, attr):
...
 
 
-if sys.version_info[0] < 3:
-    exec(dedent("""
-        class _DummyCommand(Command, object):
-            __metaclass__ = _DummyCommandMeta
-    """))
-else:
-    exec(dedent("""
+exec(dedent("""
```
saimn: The `exec` is probably no more needed.

astropy_helpers/commands/build_ext.py
```diff
@@ -255,7 +256,6 @@ def finalize_options(self):
 
             extensions = self.distribution.ext_modules
             if extensions:
-                build_py = self.get_finalized_command('build_py')
```
saimn: Is it really the same thing as `setuptools.command.build_py` ? I think it does more than that,  `setuptools.command.build_py` is just the class. But I don't know well the setuptools/distutils internals.

bsipocz:
I've tested this with the core repo here, and everything seems to be fine: https://travis-ci.org/bsipocz/astropy/builds/324278714

CHANGES.rst
```diff
@@ -4,6 +4,9 @@ astropy-helpers Changelog
 3.0 (unreleased)
 ----------------
 
+- Removing Python2 support, including 2to3. Packages wishing to keep Python2
```
astrofrog: Python2 -> Python 2

README.rst
```diff
@@ -4,11 +4,21 @@ astropy-helpers
 * Stable versions: https://pypi.org/project/astropy-helpers/
 * Development version, issue tracker: https://github.com/astropy/astropy-helpers
 
+
+.. warning::
+
+    Please note that version ``v3.0`` and later of ``astropy-helpers`` does
+    require Python >=3.5. If you wish to maintain Python2 support for your
```
astrofrog: Nitpick: I would say 'Python 3.5 or later', and Python2 -> Python 2

README.rst
```diff
@@ -4,11 +4,21 @@ astropy-helpers
...
+    package that uses ``astropy-helpers``, then do not upgrade the helpers t
+    o ``v3.0+``. We will still provide Python2.7 compatible releases on the
```
astrofrog: 'to' should not be broken up

astrofrog: Python2.7 -> Python 2.7

astropy_helpers/commands/build_ext.py
```diff
@@ -237,7 +238,7 @@ def __new__(cls, *args, **kwargs):
             # now the options being Cython.Distutils.build_ext, or the stock
             # setuptools build_ext)
             new_cls = super(build_ext, cls._final_class).__new__(
-                    cls._final_class)
+                cls._final_class)
```
astrofrog: I would not even break this line in the first place

astropy_helpers/commands/build_ext.py
```diff
@@ -303,7 +304,7 @@ def run(self):
             # Update cython_version.py if building with Cython
             try:
                 cython_version = get_pkg_version_module(
-                        packagename, fromlist=['cython_version'])[0]
+                    packagename, fromlist=['cython_version'])[0]
```
astrofrog: Same here, this would probably be ok as a single line

astropy_helpers/utils.py
```diff
@@ -293,12 +270,8 @@ def resolve_name(name):
...
 
 
-if sys.version_info[0] >= 3:
-    def iteritems(dictionary):
-        return dictionary.items()
-else:
-    def iteritems(dictionary):
-        return dictionary.iteritems()
+def iteritems(dictionary):
+    return dictionary.items()
```
astrofrog: This can be removed, and uses of this can be changed to ``.items()`` instead

astropy_helpers/commands/build_ext.py
bsipocz: the file has lines very much around 80, this would be 100, so I'll as is.

astropy_helpers/utils.py
bsipocz: Good catch, as far as I see this is not even used anywhere.

bsipocz:
Failure is unrelated to `astropy-helpers` and is a travis bug (https://github.com/travis-ci/travis-ci/issues/6307), so will open a workaround PR directly in ci-helpers.

keflavich:
My tests against python2.7 are now failing across many packages because they all attempt to install the latest astropy-helpers (though I don't know why exactly they do that).  Here's an example:
https://travis-ci.org/radio-astro-tools/radio-beam/jobs/327478267#L811

What is the recommended way to make python2.7 tests load older versions of astropy-helpers?

(some of our inline discussion on https://github.com/radio-astro-tools/spectral-cube/pull/455 is related - is that the best workaround?  Do I need to implement it everywhere?)

pllim:
@keflavich , could you please try this and see if it works?

* Remove `astropy-helpers` from conda dependencies.
* Instead of installing from `master.zip`, install from `git+https://repo_url/package.git@master`.

keflavich:
OK, I'll try that.  How is installing from git vs installing from the .zip different, though?

pllim:
Not sure. Does ZIP depend on MANIFEST? I use the git method on my package and didn't have problem (though admittedly I haven't tried with latest `astropy-helpers` yet.

pllim:
Regardless, you need to remove `astropy-helpers` as external dependency and use it as a submodule.

saimn:
I think there is a real issue with the current state of astropy_helpers because of the auto-update feature in `ah_bootstrap.py`: if you pip install from git or zip, astropy_helpers is not included so `ah_bootstrap.py` will try to install it from pip or git (I don't remember in which order etc.) and will likely install a version that is not py2 compatible.
So I guess `ah_bootstrap.py` should be modified to detect if it runs on Python 2 and install a suitable version or branch.

pllim:
@saimn , perhaps your idea is best to be submitted as a new issue so it won't be forgotten. This PR has been merged. Thanks!

PR #341, opened 2017-09-26
Allow to override build date

bmwiedemann:
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

Without this patch, the python-sunpy openSUSE package differed
for every build

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,8 @@ astropy-helpers Changelog
   strings that are in a format that ``setuptools`` expects (e.g. "1.1.3.dev0"
   instead of "1.1.3.dev"). [#330]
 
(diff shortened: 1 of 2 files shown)


astrofrog:
Thanks! I think this is a nicer solution to the problem than https://github.com/astropy/astropy-helpers/pull/146

bsipocz:
I could probably use a brief changelog entry. Could you @bmwiedemann add one for 2.0.2?

bmwiedemann:
Changelog entry was added

bsipocz:
Thanks @bmwiedemann!

PR #342, opened 2017-09-30
Use mathjax instead of imgmath

astrofrog:
We currently use mathjax on RTD but we use imgmath locally (this is to render latex equations). Using imgmath requires LaTeX to be installed in order for the docs to be built, and is slower - the astropy docs build in **5m14s** with mathjax instead of **6m40s** on my laptop, which is not negligible (these timings don't include the sphinx-gallery time since I don't have the plugin installed).

One might argue that imgmath looks nicer in some cases because it uses the real LaTeX, but this isn't important since at the end of the day what users see is the mathjax rendered LaTeX on RTD. So I am proposing we switch to also using mathjax locally and then remove the LaTeX requirements for building the docs (which we mention in http://docs.astropy.org/en/stable/install.html#building-documentation). This will then also make the local build consistent with RTD.

I also removed an unrelated obsolete comment since we do use ``sphinx.ext.viewcode`` now.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,8 @@ astropy-helpers Changelog
 
 - Mark Sphinx extensions as parallel-safe. [#344]
 
+- Switch to using mathjax instead of imgmath for local builds. [#342]
+
 2.0.1 (2017-07-28)
 ------------------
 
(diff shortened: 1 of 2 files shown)


astrofrog:
This could arguably also be milestoned as 2.0.x, so let me know what you think.

bsipocz:
Thanks @astrofrog!

PR #343, opened 2017-09-30
Update numpydoc to v0.7

astrofrog:
Fixes some issues I was seeing in parallel builds

However, before we merge this, this might be a good time to consider whether we can now create the sphinx-astropy metapackage we talked about at the coordination meeting, which would depend on numpydoc and sphinx-automodapi and the theme? Then we could remove astropy_helpers.extern


bsipocz:
I still like the idea of ``sphinx-astropy``, but also think that this update should be backported to 2.0.x, and ``sphinx-astropy`` and the new dependencies should only affect v3.0.

bsipocz:
It looks uncontroversial to me, thanks @astrofrog!

PR #344, opened 2017-09-30
Mark sphinx extensions as parallel-safe

astrofrog:
Sphinx now requires this (in order to use parallel mode). I checked over the code and as far as I can tell these should indeed be parallel safe. For sphinx-automodapi it is more complicated, and I'll copy over code from there once https://github.com/astropy/sphinx-automodapi/pull/38 is merged and I've made a release.


bsipocz:
thanks @astrofrog!

PR #346, opened 2017-10-03
Added a helper function to add OpenMP compilation flags if available

astrofrog:
This was originally based on a function by @cmccully in astroscrappy but rewritten to also test linking and running of the test code. The approach here is to make a function that modifies the extension, which seems the most convenient. Packages can then import this function in a ``setup_package.py`` and patch extensions. This function is made to fail non-catastrophically if OpenMP flags are not available (a warning is emitted and the function returns False). In the tests I've hard coded that we expect this to pass on Travis Linux and AppVeyor by default otherwise we have no way of knowing this is working for sure.

Note that the test here is a little more stringent because it checks not just compilation but also linking and running the test program, which I believe is important to make sure things run ok end-to-end.

Here's an example of how to use this in a real package: https://github.com/astropy/astropy-healpix/pull/59/files#diff-bcb5fcaf7fc1e3ab2ea2217cc624efc2

@bwinkel @wkerzendorf @cmccully - since you use -fopenmp in your packages (pycraf, tardis and astroscrappy respectively), I just thought you might be interested in this.

### DIFF
--- a/astropy_helpers/openmp_helpers.py
+++ b/astropy_helpers/openmp_helpers.py
@@ -71,9 +71,11 @@ def add_openmp_flags_if_available(extension):
     start_dir = os.path.abspath('.')
 
     if get_compiler_option() == 'msvc':
-        flag = '-openmp'
+        compile_flag = '-openmp'
+        link_flag = ''
     else:
-        flag = '-fopenmp'
+        compile_flag = '-fopenmp'
+        link_flag = '-fopenmp'
 
     try:
 
@@ -83,8 +85,8 @@ def add_openmp_flags_if_available(extension):
             f.write(CCODE)
 
         # Compile, link, and run test program
-        ccompiler.compile(['test_openmp.c'], output_dir='.', extra_postargs=[flag])
-        ccompiler.link_executable(['test_openmp.o'], 'test_openmp', extra_postargs=[flag])
+        ccompiler.compile(['test_openmp.c'], output_dir='.', extra_postargs=[compile_flag])
+        ccompiler.link_executable(['test_openmp.o'], 'test_openmp', extra_postargs=[link_flag])
         output = subprocess.check_output('./test_openmp').decode('utf-8').splitlines()
 
         if 'nthreads=' in output[0]:
@@ -109,10 +111,9 @@ def add_openmp_flags_if_available(extension):
         os.chdir(start_dir)
 
     if using_openmp:
-        log.info("Compiling Cython extension with OpenMP support (using {0} flag)".format(flag))
-        extension.extra_compile_args.append(flag)
-        if get_compiler_option() != 'msvc':
-            extension.extra_link_args.append(flag)
+        log.info("Compiling Cython extension with OpenMP support")
+        extension.extra_compile_args.append(compile_flag)
+        extension.extra_link_args.append(link_flag)
     else:
         log.warn("Cannot compile Cython extension with OpenMP, reverting to non-parallel code")
 
--- a/astropy_helpers/openmp_helpers.py
+++ b/astropy_helpers/openmp_helpers.py
@@ -33,6 +33,7 @@
 from __future__ import absolute_import, print_function
 
 import os
+import glob
 import tempfile
 import subprocess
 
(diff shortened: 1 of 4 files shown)


bsipocz:
@astrofrog - Do we want to have this on the 2.0.x branch, too so packages still supporting python2 can use it?

astrofrog:
Good point - are we going to make an exception here that features can be backported to 2.0.x? (@eteq?)

astrofrog:
An alternative is to go back on our decision to remove Python 2 support in astropy-helpers (after all, we haven't done it yet). I don't have a strong preference.

bsipocz:
> Good point - are we going to make an exception here that features can be backported to 2.0.x? (@eteq?)

as I remember we were ok with that in case of the helpers, and I'm still OK with adding any new features that doesn't break previous API for the sake of keeping the versions in sync with core.

astrofrog:
Hrm, it looks like ``register_commands`` has permanent side effects and it makes the subsequent tests fail. This is not good because some of the other tests also use ``register_commands`` so who knows what side effects those cause. I think the tests are probably overdue for refactoring.

astropy_helpers/tests/test_openmp_helpers.py
```diff
@@ -7,6 +8,7 @@
 
 IS_TRAVIS_LINUX = os.environ.get('TRAVIS_OS_NAME', None) == 'linux'
 IS_APPVEYOR = os.environ.get('APPVEYOR', None) == 'True'
+PY3_LT_35 = sys.version_info[0] == 3 and sys.version_info[1] < 5
```
bsipocz: Why not do  if sys.version_info[:2] < (3, 5), does it work for py2.7?

astrofrog: Yes it does work for Python 2.7. It's an MVCS version thingy.

bsipocz: oh, ok then, sorry for the noise

astropy_helpers/tests/test_openmp_helpers.py
```diff
@@ -0,0 +1,38 @@
...
+
+    using_openmp = add_openmp_flags_if_available(Extension('test', []))
+
+    # Make sure that on Travis (Linux) and AppVeyor OpenMP does get used (for
+    # MacOS X usually it will not work but this will depend on the compiler).
+    # Having this is useful because we'll find out if OpenMP no longer works
+    # for any reason on platforms on which it does work at the time of writing.
+    # OpenMP doesn't work on Python 3.x where x<5 on AppVeyor though.
+    if IS_TRAVIS_LINUX or (IS_APPVEYOR and not PY3_LT_35):
+        assert using_openmp
```
eteq: I would think you should invert this test so that it defaults to running if I run e.g. locally.  That is, something like:

```
if not IS_TRAVIS_OSX and not (PY3_LT_35 and IS_APPVETOR)
    assert using_openmp
```

astropy_helpers/openmp_helpers.py
```diff
@@ -0,0 +1,133 @@
...
+#     from astropy_helpers.openmp_helpers import add_openmp_flags_if_available
+#
+# then call it with a single extension as the only argument:
+#
+#     add_openmp_flags_if_available(extension)
+#
+# this will add the OpenMP flags if available.
+#
+# This file includes code heavily adapted from astroscrappy, originally released
+# under the following license:
```
eteq: Maybe this should go under the ``licenses`` directory in the root of the helpers instead of embedded in this file?

astropy_helpers/openmp_helpers.py
```diff
@@ -0,0 +1,133 @@
...
+
+        with open('test_openmp.c', 'w') as f:
+            f.write(CCODE)
+
+        os.mkdir('objects')
+
+        # Compile, link, and run test program
+        ccompiler.compile(['test_openmp.c'], output_dir='objects', extra_postargs=[compile_flag])
+        ccompiler.link_executable(glob.glob(os.path.join('objects', '*')), 'test_openmp', extra_postargs=[link_flag])
+        output = subprocess.check_output('./test_openmp').decode('utf-8').splitlines()
```
eteq: does this decode need to take ``locale.getdefaultlocale()`` or something like that? Maybe it doesn't matter, but it seems odd to be hard-coded as UTF-8?

astropy_helpers/tests/test_openmp_helpers.py
astrofrog: The issue with that is that it is not guaranteed to pass locally, e.g. if you run on MacOS X with the Apple clang then it won't pass. The idea of the current logic is that it only checks that OpenMP is being used on platforms/circumstances where we *know* it will work. What do you think?

astropy_helpers/openmp_helpers.py
astrofrog: Done

astropy_helpers/openmp_helpers.py
astrofrog: Done

astropy_helpers/tests/test_openmp_helpers.py
eteq: It still concerns me a bit there's no way to check this locally, but now that I realized it runs regardless and it's only the assert that's being checked here, I think it's ok. So all set and will merge

saimn:
Thanks for adding this, it can be useful (in MPDAF we enable OpenMP only on Linux, assuming GCC is used, so this method would be much better !).

astrofrog:
@saimn - great, I'll be interested to know if it works for you!

PR #348, opened 2017-10-10
Add fallback for when stdout locale isn't defined

astrofrog:
Small bug fix for the OpenMP helper function


astrofrog:
I cancelled the remaining OS X builds which would pass here anyway.

PR #349, opened 2017-10-10
Don't install graphviz on CI services

astrofrog:
I **think** graphviz isn't needed anymore as this was needed for the automodapi tests which are now in the sphinx-automodapi repo. This will speed up the CI a fair bit especially on Mac.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,13 +10,6 @@ os:
 # Setting sudo to false opts in to Travis-CI container-based builds.
 sudo: false
 
-# The apt packages below are needed for sphinx builds, which can no longer
-# be installed with sudo apt-get.
-addons:
-    apt:
-        packages:
-            - graphviz
-
 env:
   matrix:
     - PYTHON_VERSION=2.7 EVENT_TYPE='push pull_request cron'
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -10,25 +10,18 @@ os:
...
     - PYTHON_VERSION=3.5
     - PYTHON_VERSION=3.6 SETUPTOOLS_VERSION=dev DEBUG=True
-      CONDA_DEPENDENCIES='sphinx cython numpy packaging six appdirs'
+      CONDA_DEPENDENCIES='sphinx cython numpy packaging six appdirs coverage pytest-cov'
       EVENT_TYPE='push pull_request cron'
 
   global:
-    - CONDA_DEPENDENCIES="setuptools sphinx cython numpy"
-    - PIP_DEPENDENCIES="coveralls pytest-cov"
+    - CONDA_DEPENDENCIES="setuptools sphinx cython numpy coverage pytest-cov"
```
bsipocz: It's probably better to install these in ci-helpers when ``--cov`` is stated, isn't it?

astrofrog: @bsipocz - yes, though let's do that in a separate PR?

bsipocz: yes, and only once it's added to ci-helpers.

astrofrog:
Ah no apparently we still need graphviz for the tests. Maybe what I'll do is disable the inheritance graph. We really don't need to test that here.

.travis.yml
bsipocz: Hmm, ci-helpers currently install pytest-cov when ``--cov`` is used. I wonder whether coverage is indeed needed or will be installed as the pytest-cov dependency anyway?

astrofrog:
The Mac builds have now gone from ~12 to ~4 minutes

bsipocz:
Thank you @astrofrog!

PR #352, opened 2017-10-10
Simplifying travis matrix

bsipocz:
as ``ci-helpers`` takes care of installing pytest-cov and coveralls

(pending merging https://github.com/astropy/ci-helpers/pull/247)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,6 @@ env:
 
   global:
     - CONDA_DEPENDENCIES="setuptools sphinx cython numpy"
-    - PIP_DEPENDENCIES="coveralls pytest-cov"
     - EVENT_TYPE='push pull_request'
 
 matrix:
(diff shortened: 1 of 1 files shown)


bsipocz:
Cancelling travis while waiting for the ci-helpers merge. No need to rebase, only hit on restart once it's done.

bsipocz:
Arrg, we don't use ``SETUP_CMD`` in the travis config here so ci-helpers is not installing the packages on it's own.

However over here ``SETUP_CMD`` would not be the best name to use as we're not running setup.py test but directly pytest. Would you think we should introduce ``PYTEST_CMD`` in ci-helpers?

astrofrog:
I wonder if we might be overcomplicating this - we wouldn't be saving that many lines here as a result of the ci-helpers change. Maybe it's not the end of the world to have the coverage dependencies here?

PR #360, opened 2018-01-13
Add python_requires to setup.py

astrofrog:
Just in case anyone tries to pip install astropy-helpers, we should declare what Python versions it works with using ``python_requires``

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ env:
       EVENT_TYPE='push pull_request cron'
 
   global:
+    - SETUPTOOLS_VERSION=27
     - CONDA_DEPENDENCIES="setuptools sphinx cython numpy"
     - PIP_DEPENDENCIES="coveralls pytest-cov"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -18,6 +18,7 @@ env:
       EVENT_TYPE='push pull_request cron'
 
   global:
+    - SETUPTOOLS_VERSION=27
```
bsipocz: strictly speaking we could use a maximum version number, but I'm fine with pinning it to be this precise.

astrofrog: Good point - if you are happy leaving it as is, let's do that, since it's (hopefully) only temporary.

bsipocz: sure, I suspect it's available for all supported python versions (for the backport), if not we can always change it there.

astrofrog:
The Travis failures are unrelated and are to do with the latest developer versions of setuptools and sphinx

Issue #359, opened 2018-01-11
Possible issue with ah_bootstrap and Python 2


saimn:
From https://github.com/astropy/astropy-helpers/pull/340#issuecomment-356805492 (and following comments), it is likely that the removal of Python 2 support is the root of the issue, because of the automatic download/installation of astropy_helpers by ah_bootstrap. I think ah_bootstrap needs to be updated to install a Python 2 compatible version of astropy_helpers in this case ?

astrofrog:
Ah we should check if astropy-helpers has ``python_requires >= 3.5`` to prevent it from getting downloaded (just checked, it doesn't so we definitely need to add).

But anyway, how is astropy-helpers 3.0rc1 getting downloaded automatically since it's just an RC?

saimn:
Because it uses easy_install, which does not support `--pre` (or `python_requires`):
```
testtmp ❯ easy_install astropy_helpers
Searching for astropy_helpers
Reading https://pypi.python.org/simple/astropy_helpers/
Downloading https://pypi.python.org/packages/c4/0f/e577377e118c0a4c2de3d9c143c9e383053b9dc271236b53026957dae3a4/astropy-helpers-3.0rc1.tar.gz#md5=93f3f47e212ab92ea961556912ce6b6e
Best match: astropy-helpers 3.0rc1
Processing astropy-helpers-3.0rc1.tar.gz
Writing /tmp/simon/easy_install-iu9qb341/astropy-helpers-3.0rc1/setup.cfg
Running astropy-helpers-3.0rc1/setup.py -q bdist_egg --dist-dir /tmp/simon/easy_install-iu9qb341/astropy-helpers-3.0rc1/egg-dist-tmp-9yr20wuu
/tmp/simon/easy_install-iu9qb341/astropy-helpers-3.0rc1/astropy_helpers/setup_helpers.py:361: AstropyDeprecationWarning: Use of the exclude parameter is no longer supported since it does not work as expected. Use add_exclude_packages instead. Note that it must be called prior to any other calls from setup helpers.
  AstropyDeprecationWarning)
/tmp/simon/easy_install-iu9qb341/astropy-helpers-3.0rc1/astropy_helpers/setup_helpers.py:672: AstropyDeprecationWarning: Use of the exclude parameter is no longer supported since it does not work as expected. Use add_exclude_packages instead. Note that it must be called prior to any other calls from setup helpers.
  AstropyDeprecationWarning)
[...]

Issue #361, opened 2018-01-15
Tests broken with setuptools 36 and above


astrofrog:
Since the following commit in setuptools: https://github.com/zhangshuai930822/python-setuptools/commit/472c79f95cc41a3e85d6b212347d6b006c9c3c26 - some of the tests in astropy-helpers are failing, e.g.:

```
=============================================================== test session starts ================================================================
platform darwin -- Python 3.6.4, pytest-3.3.2, py-1.5.2, pluggy-0.6.0
rootdir: /Users/tom/Dropbox/Code/Astropy/astropy-helpers, inifile: setup.cfg
plugins: hypothesis-3.38.5
collected 28 items                                                                                                                                 

astropy_helpers/tests/test_ah_bootstrap.py F                                                                                                 [100%]

===================================================================== FAILURES =====================================================================
___________________________________________________________ test_bootstrap_from_archive ____________________________________________________________

    @contextlib.contextmanager
    def save_modules():
        """
        Context in which imported modules are saved.
    
        Translates exceptions internal to the context into the equivalent exception
        outside the context.
        """
        saved = sys.modules.copy()
        with ExceptionSaver() as saved_exc:
>           yield saved

/Users/tom/tmp/setuptools/setuptools/sandbox.py:158: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

setup_dir = '/private/var/folders/zy/t1l3sx310d3d6p0kyxqzlrnr0000gr/T/pytest-of-tom/pytest-27/test_bootstrap_from_archive0/orig'

    @contextlib.contextmanager
    def setup_context(setup_dir):
        temp_dir = os.path.join(setup_dir, 'temp')
        with save_pkg_resources_state():
            with save_modules():
                hide_setuptools()
                with save_path():
                    with save_argv():
                        with override_temp(temp_dir):
                            with pushd(setup_dir):
                                # ensure setuptools commands are available
                                __import__('setuptools')
>                               yield

/Users/tom/tmp/setuptools/setuptools/sandbox.py:199: 
[...]

PR #362, opened 2018-01-16
Fix fetching of astropy-helpers on Python 2

astrofrog:
As reported on https://github.com/astropy/astropy-helpers/issues/359 - the issue is that if the astropy-helpers submodule is not present or not checked out on Python 2, ``ah_bootstrap.py`` will then fetch the absolute latest version of astropy-helpers. This PR introduces a strict upper limit to the versions that will be fetched.

I considered another solution that would re-use ``_do_upgrade`` to only get the latest bugfix release, but when the submodule is not checked out, ``ah_bootstrap.py`` is not really tied to a specific astropy-helpers version - in fact maybe the reason some people don't include the submodule is precisely to not have to update astropy-helpers as often. So I think the fix in this PR is the right way to go.

This should be backported to v2.0.x - by the way this does raise the question of how we will update the astropy-helpers in different repos with these releases - how do we know whether to use the latest v2.0.x release or the v3.0.0 release? (maybe we can do some heuristics based on ``setup.py`` and e.g. ``.travis.yml``?)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,7 +17,7 @@ astropy-helpers Changelog
 2.0.3 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Make sure that astropy-helpers 3.x.x is not downloaded on Python 2. [#362, #363]
 
 
 2.0.2 (2017-10-13)
--- a/ah_bootstrap.py
+++ b/ah_bootstrap.py
@@ -139,6 +139,7 @@ from distutils.debug import DEBUG
 # TODO: Maybe enable checking for a specific version of astropy_helpers?
 DIST_NAME = 'astropy-helpers'
 PACKAGE_NAME = 'astropy_helpers'
+UPPER_VERSION_EXCLUSIVE = None
 
 # Defaults for other options
 DOWNLOAD_IF_NEEDED = True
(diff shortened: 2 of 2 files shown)


astrofrog:
There is a separate PR to v2.0.x: https://github.com/astropy/astropy-helpers/pull/363

bsipocz:
> This should be backported to v2.0.x - by the way this does raise the question of how we will update the astropy-helpers in different repos with these releases - how do we know whether to use the latest v2.0.x release or the v3.0.0 release? (maybe we can do some heuristics based on setup.py and e.g. .travis.yml?)

What I had in mind is that we keep doing the v2.0.x auto-updates, but for 3.0.0, we open an issue in their issue tracker to emphasize that it's python3 only and ask them to opt in when they are ready for it. We can run then the update based on a separate list, not the full affiliate registry.

astrofrog:
Ah right yes, good idea - in fact we could just host the list of packages to update in the tools repo with for each package the major version of astropy-helpers to use - we could then include a note in the auto-update PR (rather than opening a separate issue) that they can do a PR to that list to opt in?

bsipocz:
@astrofrog - Yes, that's a good idea.

bsipocz:
Merging now. The failures are related to #361 (setuptools) and the sphinx-dev issue will be fixed in #365. On slack we agreed to temporarily allow to fail the setuptools >27 tests until #361 is figured out, I'll do that separately.

bsipocz:
Thanks @astrofrog!

Issue #338, opened 2017-09-25
Split theme out into separate package


astrofrog:
At the coordination meeting in NY there were no objections to splitting the sphinx theme out into a separate package. I think that for consistency with other plugin names like ``sphinx-gallery``, ``sphinx-automodapi``, and theme package names like ``sphinx-rtd-theme`` this package should be called ``sphinx-astropy-theme``. We also discussed that the cleanest way to get people to install all the plugins required would be to make a ``sphinx-astropy`` metapackage.

I've stated a repo here for the theme:

https://github.com/astropy/sphinx-astropy-theme

Does anyone volunteer to take a stab at this?

Cadair:
https://github.com/Cadair/astropy-sphinx-theme

Cadair:
I did it a long time ago...

astrofrog:
Great! Could you open a PR to the new repo to add the latest files? (since things might have changed a little since then)

saimn:
Could be worth doing a proper extraction from git to keep the history ? It's not hard to do, and always good to know when and by who things were done.

astrofrog:
@saimn @Cadair - I redid this preserving all commits and including some of @Cadair's new commits: https://github.com/astropy/sphinx-astropy-theme - does this seem ok to you? The filtering process only kept one non-trivial merge commit, but I don't think we need to actually keep the other merge commits?

Cadair:
Lgtm!

saimn:
Sounds great !

PR #367, opened 2018-01-19
Remove Sphinx Astropy theme

astrofrog:
As discussed in https://github.com/astropy/astropy-helpers/issues/338, I've moved the Astropy theme out into a standalone repository:

https://github.com/astropy/sphinx-astropy-theme

Now if people try and build the docs and the theme is not installed, they will see the following error:

screen shot 2018-01-19 at 18 11 12

Note however the message in blue which is not a warning (and will therefore not trigger issues for people trying to build without warnings) but which should be prominent enough to catch their attention.

At the moment there aren't any releases of sphinx-astropy-theme, but if people are in favor of this PR, I can go ahead and do a release before we merge this (I was thinking of starting versioning the theme starting at 1.0 since we've been using it for a while).


pllim:
From a package maintainer POV, I would like switch-over instructions; e.g., "If you update to this new version of astropy-helpers, you need to do this, this, and this to get the new theme and get rid of old code."

astrofrog:
I guess the instructions are just 'if you update to this new version of astropy-helpers, you need to install sphinx-astropy-theme' manually :) It's a good point that maybe we should keep a list of things like that to mention in the PR that auto-updates astropy-helpers though?

pllim:
Perhaps an update to the package template is also in order?

astrofrog:
@pllim - yes but let's wait until more work is done on astropy-helpers, because I think we also want to unbundle numpydoc and sphinx-automodapi, which means we really should just have a sphinx-astropy package that users can install instead of all the different separate packages.

astrofrog:
(the tests are failing because I need to install the theme for the CI to run)

setup.py
```diff
@@ -47,5 +47,6 @@
     cmdclass=cmdclass,
     python_requires='>=3.5',
     zip_safe=False,
+    install_requires=['sphinx-astropy-theme'],
```
Cadair: You don't actually *need* the theme to use helpers... I can understand why you would want to keep it though.

Cadair: You could make the `conf.py` fall back to the default sphinx `alablaster` theme if the astropy one isn't installed.

astrofrog: I don't think we should have a fallback as I think for now users won't want the theme to suddenly change, and it's easier to tell users how to install the theme if there is an error.

astrofrog:
So are we ok with an error if the theme isn't installed? I don't think there is a way to auto-install as if the submodule is present I don't think astropy-helpers is actually installed so ``install_requires`` in the helpers isn't honored.

Alternatively we could do as @Cadair suggests and have a fallback theme.

At the end of the day, the only people building the docs are people making contributions, so I think they might be ok with installing a separate package. By the time 3.1 comes around we might be asking them to just install ``sphinx-astropy`` to build the docs anyway (not just the theme).

astrofrog:
Note that we can also wait until sphinx-astropy is in place before considering how to deal with it here. I'm going to be working on that over the weekend (there's no rush to merge)

astrofrog:
Closing in favor of #368

Issue #339, opened 2017-09-25
Remove bundled sphinx-automodapi


astrofrog:
As discussed in https://github.com/astropy/astropy-helpers/issues/338, we agreed at the NY meeting that we should split sphinx code out into standalone packages. This was already done for sphinx-automodapi but we should unbundle it and instead make it a dependency of the sphinx-astropy metapackage (once this exists).

In fact, we should check if the remaining sphinx code which is not generically useful could just be moved to ``sphinx-astropy`` (which then wouldn't just be a meta-package)

Issue #327, opened 2017-07-21
Intersphinx failure with NumPy


MSeifert04:
In `ccdproc` we had a sphinx build failure because the NumPy doc server was down (Link to build: https://travis-ci.org/astropy/ccdproc/jobs/255026568#L677)

[@bsipocz mentioned that that shouldn't happen](https://github.com/astropy/ccdproc/pull/515#issuecomment-316360114)

> Hmm, this is unfortunate. The scipy and numpy intersphinx are actually picked up from the stsci server, so that warnings should have been caught and never have made the build fail. Could you please open an issue for it in astropy-helpers?

So here's the issue :)

bsipocz:
Thanks for opening. We need to make sure that this warning is ignored when the intersphinx is picked up from an alternative source. 

Probably it can be addressed upstream in sphinx, but if not we need to deal with it ourselves here.

Issue #276, opened 2016-12-09
automodapi inheritance diagram self-looping arrow


pllim:
Most people probably won't encounter this but it does happen when you have a subclass that shares the same name as its parent class. Example:

```python
from astropy.modeling import models as _models

class Box1D(_models.Box1D):
    """Extends the original Box1D with some special magic."""
    pass
```

The resultant inheritance diagram using `automodapi` makes a loop of the class on itself. It is harmless but appears quite funny and potentially confusing. I don't even know if a fix is possible or necessary. Just want to point it out.

![untitled](https://cloud.githubusercontent.com/assets/2090236/21055528/b5a5831c-bdff-11e6-8308-85eaa6dd509f.png)

eteq:
Oh interesting... I'm not sure this can be fixed exactly the way you might expect, because the underlying diagram code uses graphviz, which keys on names.

Before I try to fix, if there *isn't* a way to have them show the same name, what do you think would make sense @pllim ?  Maybe (something like ``Box1D (#2)``?

pllim:
GitHub has spoiled the `#` for me as I associate it with issue number now. I am not sure what would be a less confusing representation. `Box1D (2)` (without the `#` maybe) might just be as good as any?

pllim:
Or maybe just leave it as `Box1D -> Box1D`? Since when you mouse over, you will get more info about where the class is coming from anyway, right?

eteq:
@pllim - don't think your second option here is possible because of the origin of ths issue (that graphviz keys on names.)  But it *might* be that "Box1D" and "Box1D " (note the space in the second one) are considered different?  That will acheive a very similar appearence, at least...

pllim:
@eteq , sure, I'm not picky on this.

Issue #233, opened 2016-05-08
Warning from numpydoc doesn't trigger Sphinx fail.


MSeifert04:
The warning emitted in [NumpyDocString](https://github.com/astropy/astropy-helpers/blob/master/astropy_helpers/sphinx/ext/docscrape.py#L120) isn't catched by sphinx so it passes the travis CI checks - though the section isn't even rendered in the output. I think these warnings should fail sphinx-builds.

A bit more information (not much) can be found in a recent PR to fix such a warning: https://github.com/astropy/astropy/pull/4852

astrofrog:
Can someone investigate this to see if it is still an issue?

bsipocz:
This should have been fixed on the numpydoc side,  at least we had tons of failures when they switched on failing on these warnings since 0.7?. 
See e.g. https://github.com/numpy/numpydoc/issues/115 and some of the v2.0.2 helpers update PRs.

PR #368, opened 2018-01-29
Remove all sphinx components

astrofrog:
### Overview

This is a more extensive alternative to #367, and it:

* Removes the theme which is in https://github.com/astropy/astropy-sphinx-theme (and on PyPI)
* Removes the bundled sphinx-automodapi and numpydoc
* Removes the rest of ``astropy_helpers.sphinx`` to https://github.com/astropy/sphinx-astropy

It does **not** remove ``build_docs`` though.

Note that the astropy-sphinx-theme and sphinx-astropy repositories contain the full history of the files in those repos, including their origins in the astropy core package. I am publishing the scripts I used to do all this here: https://github.com/astrofrog/git-workflows (note that this isn't up to date yet as I have some more changes to push).

Note that sphinx-astropy is also a meta-package in that it includes dependencies on astropy-sphinx-theme, sphinx-automodapi, numpydoc, and sphinx-gallery. I haven't published sphinx-astropy yet as I want to make sure we have a chance to discuss this here first.

### Backward-compatibility

These changes will require one change in astropy core and affiliated packages (when they update to astropy-helpers next), which to change this line:

https://github.com/astropy/astropy/blob/master/docs/conf.py#L48

from:

```python
from astropy_helpers.sphinx.conf import *
```

to

```python
from sphinx_astropy.conf import *
```

Note that we can make this change automatically when updating astropy-helpers. We could optionally add a nicer error message:

```python
try:
    from sphinx_astropy.conf import *
except ImportError:
    print('Note that the documentation for this package requires the sphinx-astropy package to be installed')
    sys.exit(1)
```

This has the benefit of being super explicit in terms of what is required, which is an improvement over just the logging warning in #367.

See https://github.com/astropy/astropy/pull/7139 for an example implementation in the core package.

I don't *think* we need to add a compatibility layer to astropy-helpers since in principle packages will be able to update their ``conf.py`` at the same time as the astropy-helpers submodule.

### Steps forward

If we agree on this, these are the steps ahead:

* [x] Merge https://github.com/astropy/sphinx-astropy/pull/2
* [x] Do a release of the sphinx-astropy package
* [x] Update this PR to use the released sphinx-astropy package
* [x] Determine if the Travis matrix can be simplified (fewer sphinx versions)
* [ ] Open a PR to astropy core to use sphinx-astropy instead of astropy-helpers in conf.py, and mention dependency on sphinx-astropy to build the docs (https://github.com/astropy/astropy/pull/7139)
* [ ] Open a PR to the package template to use sphinx-astropy instead of astropy-helpers in conf.py

Fixes https://github.com/astropy/astropy-helpers/issues/338
Fixes https://github.com/astropy/astropy-helpers/issues/339

The following issue should probably be moved to sphinx-astropy, or at least resolved there then closed here: https://github.com/astropy/astropy-helpers/issues/270, https://github.com/astropy/astropy-helpers/issues/195, https://github.com/astropy/astropy-helpers/issues/211, https://github.com/astropy/astropy-helpers/issues/233, https://github.com/astropy/astropy-helpers/issues/320, https://github.com/astropy/astropy-helpers/issues/327

The following issue should probably be in sphinx-automodapi; https://github.com/astropy/astropy-helpers/issues/276


bsipocz:
> Open a PR to the package template to use sphinx-astropy instead of astropy-helpers in conf.py

@astrofrog - Just to double check, to use `sphinx-astropy` a package doesn't have to use this 3.1 version of the helpers, right?

.travis.yml
```diff
@@ -60,6 +60,9 @@ before_script:
     - git config --global user.name "A U Thor"
     - git config --global user.email "author@example.com"
 
+    # TEMP: replace by sphinx-astropy in PIP_DEPENDENCIES once installed
```
bsipocz: once installed or once released?

CHANGES.rst
```diff
@@ -6,6 +6,10 @@ astropy-helpers Changelog
 
 - Removing deprecated test_helpers.py file. [#369]
 
+- Remove all sphinx components from astropy-helpers. These are now replaced
+  by the sphinx-astropy package in conjunction with the astropy-theme-sphinx,
```
bsipocz: please double backtick the package names

astropy_helpers/commands/build_sphinx.py
```diff
@@ -120,7 +120,10 @@ def run(self):
         build_cmd_path = os.path.abspath(build_cmd.build_lib)
 
         ah_importer = pkgutil.get_importer('astropy_helpers')
-        ah_path = os.path.abspath(ah_importer.path)
+        if ah_importer is None:
```
bsipocz: what prompted this change?

bsipocz:
I really like this refactoring, but have a few questions left.

Do we plan to pin sphinx-astropy's version in the helpers releases? I just wonder about scenarios when it's not safe or desired by affiliated packages to upgrade to the latest sphinx-astropy. Since having this factored out, the overall complexity of the helpers package hierarchy is growing that may cause further confusion.

.travis.yml
astrofrog: Yep, good catch! I'll fix it before we merge.

astropy_helpers/commands/build_sphinx.py
astrofrog: I think this was actually related to trying to build the docs with ``python setup.py build_docs --use-system-astropy-helpers`` so isn't quite related to this PR but I found it while trying to use this PR

astrofrog:
> @astrofrog - Just to double check, to use sphinx-astropy a package doesn't have to use this 3.1 version of the helpers, right?

No, you can use sphinx-astropy without ever installing the helpers.

> Do we plan to pin sphinx-astropy's version in the helpers releases? I just wonder about scenarios when it's not safe or desired by affiliated packages to upgrade to the latest sphinx-astropy. Since having this factored out, the overall complexity of the helpers package hierarchy is growing that may cause further confusion.

As far as I know there is no way to 'pin' sphinx-helpers in astropy-helpers as such, since astropy-helpers is never actually installed (so ``install_requires`` is not used). What we'll need to do is add sphinx-astropy to the ``.travis.yml`` of packages that test the docs build, and that's where the pinning would need to happen. But I'm not sure if we need to pin it, no more than we need to pin e.g. numpydoc or sphinx-automodapi. The great thing about having sphinx-astropy as a standalone package is we can easily do quick bugfix releases if downstream issues occur.

bsipocz:
> But I'm not sure if we need to pin it, no more than we need to pin e.g. numpydoc or sphinx-automodapi. The great thing about having sphinx-astropy as a standalone package is we can easily do quick bugfix releases if downstream issues occur.

But those were bundled into the helpers so far, so we had some control over what to include. You're probably right though that we can have a hot fix when any issues come up. I would just like to avoid situations when the random xy package start complaining with some weird looking error that comes up at a random occasion and the solution will be that we have updated something in sphinx-astropy (but since the packages are low traffic ones, that update might have happened a while ago).

astrofrog:
This does raise a good point, which might have come up anyway even without this split - what if we want to change the default configuration some day? One solution since we are updating the imports anyway would be to update to versioned imports:

```
from sphinx_astropy.conf.v1 import *
```

this then allows us to make significant changes to the default configuration if we want in future (without requiring packages pin to a specific version).

bsipocz:
Yes, the approach could work, and with that we may also want to consider to backport this refactoring to the 2.0.x branch, so anyone still supporting python2 could use sphinx-astropy.

Or it's already python3 only?

astrofrog:
> Yes, the approach could work, and with that we may also want to consider to backport this refactoring to the 2.0.x branch, so anyone still supporting python2 could use sphinx-astropy.

Anyone can use sphinx-astropy irrespective of astropy-helpers version. This PR just removes all the code from astropy-helpers but that's just cleanup - you can use sphinx-astropy even if you are on v2.0.x, except if sphinx-astropy isn't compatible with Python 2 (which I need to check).

Basically if you import ``sphinx_astropy.conf``, then the code in astropy-helpers gets ignored.

eteq:
Generally I'm 👍 to this, @astrofrog. However, two concerns:

First: Am I understanding right that you are *not* then planning to package all the new "requirements" as helpers dependencies? If so that seems problematic because one of the main selling points of the helpers is that it's a "one stop shop" and the user of it needn't worry too much about managing all their dependencies.  Is there any reason *not* to have the helpers setup.py now have a ``requires=[..., 'sphinx-astropy']``? 

Second:

> These changes will require one change in astropy core and affiliated packages

Why not, at least for a time, leave the existing import in place but have it simply be a placeholder ``astropy_helpers/sphinx/conf.py`` that does
```
from sphinx_astropy.conf import * 

warn(DeprecationWarning('Need to do the new fancy helpers thingie!'))
```
Or similar?  That lets helpers users (i.e. affiliated package devs) know that this is coming, but gives them a window to do the helpes update *without* it breaking immediately?

bsipocz:
> Or similar? That lets helpers users (i.e. affiliated package devs) know that this is coming, but gives them a window to do the helpes update without it breaking immediately?

we can do that deprecation you suggest, but I would rather include the conf change in the same automated PR that updates for the 3.1 helpers that raising any deprecation warnings to the users. On the other hand, not deprecating but having that import in the helpers sounds like a good solution to me, too.

astrofrog:
> First: Am I understanding right that you are not then planning to package all the new "requirements" as helpers dependencies? If so that seems problematic because one of the main selling points of the helpers is that it's a "one stop shop" and the user of it needn't worry too much about managing all their dependencies. Is there any reason not to have the helpers setup.py now have a requires=[..., 'sphinx-astropy']?

We can add it to ``install_requires`` - however note that for packages that make use of the astropy-helpers as a *submodule* this might not be honored because then the helpers are not really installed as such. But I can definitely add it for cases where it is installed.

There are a couple of options here as far as I can see - we could either modify ``build_docs`` to automatically install ``sphinx-astropy`` temporarily when running ``build_docs`` (like we do for ``pytest-astropy`` when running the tests) or we could modify ``ah_bootstrap.py`` to automatically install sphinx-astropy also temporarily while running setup commands (e.g. build_docs). I can investigate both options to see what would work.

> Why not, at least for a time, leave the existing import in place but have it simply be a placeholder astropy_helpers/sphinx/conf.py

Sounds good, that's easy to do!

astrofrog:
@eteq - I've implemented the auto-install and the warning. The auto-install can't be tested completely yet as I need to do a release of sphinx-astropy first, but I tested it by using another package. Would something like this work for you? (this is like what we do for running tests). If you are happy with it I could do a release of sphinx-astropy to check it works properly?

astropy_helpers/sphinx/conf.py
```diff
@@ -0,0 +1,5 @@
+import warnings
+
+from sphinx_astropy.conf import * 
+
+warnings.warn("Note that astropy_helpers.sphinx.conf is deprecated - use sphinx_astropy.conf instead")
```
eteq: (minor) maybe "use sphinx_astropy.conf in the sphinx-astropy package"?  Just so a random user who pays no attention this stuff knows what to google...

eteq:
One thought which is not strictly related to this PR: should we update the astropy docs to explicitly say somewhere "got to the sphinx-astropy repo to get help/report issues/etc on the sphinx machinery"?

astrofrog:
> One thought which is not strictly related to this PR: should we update the astropy docs to explicitly say somewhere "got to the sphinx-astropy repo to get help/report issues/etc on the sphinx machinery"?

I have mixed feelings about this - if there are issues with building the astropy docs, it could be due to an issue in astropy itself, the helpers, or any of the extensions, or sphinx itself. People who understand the build process will likely report issues in the right place, and otherwise for users I think it's fine if they just open an issue in the main repository and we can dispatch as needed.

eteq:
> This will not auto-install Sphinx because Sphinx itself has to be installed in the first place for build_docs to even run

That was a typo, my bad - I meant the sphinx *plugin* installs.  So I'm happy as far as that is concerned.

> I have mixed feelings about this - if there are issues with building the astropy docs, it could be due to an issue in astropy itself, the helpers, or any of the extensions, or sphinx itself. People who understand the build process will likely report issues in the right place, and otherwise for users I think it's fine if they just open an issue in the main repository and we can dispatch as needed.

Hmm, that worries me for *us* because it's hard to keep a 30,000 foot view.  And I think some users get turned off by the "thanks for your input, but that should go in the `astropy-sphinx` tracker, so I'm closing this and moving it there.  Mostly just thinking aloud though - I think your answer here is reasonable and there isn't an obvious better solution...

Anyway, all of these are not issues, so I'm going to go ahead and merge.

PR #370, opened 2018-02-06
Support dotted package names as namespace packages

lpsinger:
I am making a new project using the Astropy affiliated package where the root of the package lives inside another namespace package. For concreteness, the package is called [`ligo.skymap` or `ligo-skymap`](http://git.ligo.org/leo-singer/ligo.skymap), the top-level package is called `ligo.skymap`, and it lives inside a namespace shared by a number of other packages [`ligo.lvalert`](https://git.ligo.org/lscsoft/lvalert), [`ligo.gracedb`](https://git.ligo.org/lscsoft/gracedb-client).

However, this does not work with astropy_helpers because the function `generate_version_py` in astropy_helpers expects that the package name to be the same as the directory containing `version.py`.

This patch allows the package name to contain dots, which are interpreted as the fully-qualified name of a module. The dots are translated to path separators before writing the `version.py` script.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,7 +22,8 @@ astropy-helpers Changelog
 2.0.4 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Support dotted package names as namespace packages in generate_version_py.
+  [#370]
 
 
 2.0.3 (2018-01-20)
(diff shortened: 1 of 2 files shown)


astrofrog:
@lpsinger - thanks! Can you add a new 2.0.4 section to the changelog and include a note there about the change?

CHANGES.rst
```diff
@@ -22,7 +22,7 @@ astropy-helpers Changelog
 2.0.4 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Support dotted package names as namespace packages. [#370]
```
astrofrog: Maybe say ``in generate_version_py``

bsipocz: OK, will do it before merge, but now rather wait for the CI to finish.

bsipocz:
thanks @lpsinger!

PR #372, opened 2018-02-08
Fix compatibility with the latest versions of setuptools

astrofrog:
The first change is to call fetch_build_eggs explicitly as in recent versions of setuptools the setup_requires does not appear to be respected. The second change is to explicitly parse the setup.cfg file as this appears to not be done inside Distribution anymore but at a higher level.

Fixes https://github.com/astropy/astropy-helpers/issues/361

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@ astropy-helpers Changelog
 3.0 (unreleased)
 ----------------
 
+- Fix compatibility with the latest version of setuptools. [#372]
+
 - Removing Python 2 support, including 2to3. Packages wishing to keep Python
   2 support should NOT update to this version. [#340]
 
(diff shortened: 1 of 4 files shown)


astrofrog:
I've milestoned this as v2.0.4 as I think this is not critical for v3.0 but we should probably include in v2.0.4 and v3.0.1. I think this doesn't affect packages which use astropy-helpers as a submodule, meaning most of them.

CHANGES.rst
```diff
@@ -18,11 +18,10 @@ astropy-helpers Changelog
 - Removing unused 'register' command since packages should be uploaded
   with twine and get registered automatically. [#332]
 
-
 2.0.4 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Fix compatibility with the latest version of setuptools. [#372]
```
bsipocz: maybe give the version number here

Issue #282, opened 2016-12-19
Remove or update tox.ini


bsipocz:
I believe we don't use it any more, and thus ``tox.ini`` contains rather obsolete versions. 

Should we remove it altogether or update the versions in it?

cc @eteq

PR #373, opened 2018-02-08
Remove outdated tox.ini since we don't use it

astrofrog:
Fixes https://github.com/astropy/astropy-helpers/issues/282


astrofrog:
This doesn't have to be backported so can be re-milestoned as 3.1 if needed.

PR #375, opened 2018-02-20
Fix segmentation faults

astrofrog:
In various packages using astropy-helpers it is possible in some cases to trigger a segmentation fault when the submodule is first initialized. This is described in more detail in https://github.com/cython/cython/issues/2104

The bottom line is that this seems to be due to the interaction between setuptools' ``run_setup`` command and Cython, and the following script fails on various combinations of Cython and setuptools:

```python
import os

with open('test.pyx', 'w') as f:
    f.write('cimport cython\n')

with open('test.py', 'w') as f:
    f.write('import Cython.Compiler.Main')

from setuptools.sandbox import run_setup
run_setup(os.path.join(os.path.abspath('.'), 'test.py'), ['egg_info'])

from Cython.Build import cythonize
cythonize('test.pyx')
```

This PR updates ``ah_bootstrap.py`` to use ``subprocess`` instead of ``run_setup``, which works just as well. The test suite was also updated to use ``subprocess`` otherwise the use of ``subprocess`` in ``ah_bootstrap.py`` raised a sandbox violation error.

In any case, I'm actually happier with using subprocess since this keeps things more self-contained, especially when it comes to the tests.

For anyone reviewing this, most of the changes are to the tests, but the main change to check is the one in ``ah_bootstrap.py``.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,8 +34,7 @@ astropy-helpers Changelog
 ------------------
 
 - Fix segmentation faults that occurred when the astropy-helpers submodule
-  was first initialized and packages using astropy-helpers contained
-  Cython code. [#375]
+  was first initialized in packages that also contained Cython code. [#375]
 
 2.0.4 (2018-02-09)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -33,8 +33,9 @@ astropy-helpers Changelog
 2.0.5 (unreleased)
 ------------------
 
-- Nothing changed yet.
-
+- Fix segmentation faults that occurred when the astropy-helpers submodule
+  was first initialized and packages using astropy-helpers contained
+  Cython code. [#375]
 
 2.0.4 (2018-02-09)
 ------------------
--- a/astropy_helpers/conftest.py
+++ b/astropy_helpers/conftest.py
@@ -1,0 +1,55 @@
+import os
+import glob
+
+try:
+    from coverage import CoverageData
+except ImportError:
+    HAS_COVERAGE = False
+else:
+    HAS_COVERAGE = True
+
+# Since we run many of the tests in sub-processes, we need to collect coverage
+# data inside each subprocess and then combine it into a single .coverage file.
+# To do this we set up a list which run_setup appends coverage objects to.
+
+if HAS_COVERAGE:
+    SUBPROCESS_COVERAGE = []
+
+
+def pytest_configure(config):
+    if HAS_COVERAGE:
+        SUBPROCESS_COVERAGE.clear()
+
+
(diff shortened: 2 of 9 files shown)


CHANGES.rst
```diff
@@ -29,8 +29,9 @@ astropy-helpers Changelog
 2.0.5 (unreleased)
 ------------------
 
-- Nothing changed yet.
-
+- Fix segmentation faults that occurred when the astropy-helpers submodule
+  was first initialized and packages using astropy-helpers contained
```
eteq: (nitpicking) maybe :"was first initialized *in* packages"?

astropy_helpers/tests/__init__.py
```diff
@@ -39,18 +37,24 @@ def run_cmd(cmd, args, path=None, raise_error=True):
...
-        return sandbox.run_setup(*args, **kwargs)
-    finally:
-        import importlib
-        importlib.invalidate_caches()
+    p = sp.Popen([sys.executable, setup_script] + list(args), cwd=path,
+                 stdout=sp.PIPE, stderr=sp.PIPE)
+    stdout, stderr = p.communicate()
+
+    sys.stdout.write(stdout.decode('utf-8'))
+    sys.stderr.write(stderr.decode('utf-8'))
```
eteq: Isn't this subtly different in that the process has to *finish* before any output is visible?  So any time this is used the process will appear to hang?  Why not instead pass stdout/stderr in to the Popen?

astrofrog: Just to clarify, this ``run_setup`` is used **only** in the tests, and the only ``run_setup`` call in ``ah_bootstrap.py`` was in a ``silence`` context manager (it was for the helpers, not the actual package being built). The ``run_setup`` function is not used in the astropy build process.

astrofrog:
This now includes some extra code in the testing framework to deal with properly measuring the coverage in the subprocesses. This is not ideal but at least it only adds a bit of complication to the test suite but not the main astropy-helpers code.

astropy_helpers/conftest.py
```diff
@@ -0,0 +1,55 @@
+import os
```
eteq: This seems to be re-introducing pytest-specific machinery just after we've removed it to other packages.  Is that desirable?

And even if we are OK with that, there should probably be something at the top that amounts to "this is to make coverage work, *not* meant for general pytest plugin stuff"?

astrofrog: Just to clarify, this isn't really a plugin, and it really is very specific to astropy-helpers, so I'll just add a comment to this effect.

astrofrog: @eteq - fixed

PR #379, opened 2018-02-23
Fix package exclusion behavior in our own setup.py

drdavella:
This is somewhat embarrassing, but our own `setup.py` within `astropy_helpers` was using deprecated behavior. This PR fixes it.

This should address the issue reported here, although it will need to be backported to `v2.0.5` if it's not too late: https://github.com/gammapy/gammapy/pull/1264#issuecomment-368108877

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -39,7 +39,7 @@ astropy-helpers Changelog
 2.0.6 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Avoid deprecation warning due to ``exclude=`` keyword in ``setup.py``. [#379]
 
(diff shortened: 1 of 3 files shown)


astrofrog:
@drdavella - can you add a changelog entry for 2.0.6? (unfortunately 2.0.5 is already done)

astrofrog:
I just added a changelog entry

bsipocz:
I wonder how we never noticed this. would there be a way to test for similar issues?

astrofrog:
@bsipocz - I think it was probably just raising deprecation warnings in the Travis logs

PR #381, opened 2018-03-12
Don't capture stderr with -w

Cadair:
The line that the subprocess checks for is in stdout, so we don't need to capture stderr, which allows it to be piped away to hide warnings in the build.

---

fwiw, I am using this in sunpy/sunpy#2553 where I modify the sphinx logger to send sphinx warnings to stdout not stderr, so that I can pipe stderr to `/dev/null` and massively reduce the number of output lines on CI to make it easier for people to find the actual sphinx warnings that are making their builds fail.

### DIFF
--- a/astropy_helpers/commands/build_sphinx.py
+++ b/astropy_helpers/commands/build_sphinx.py
@@ -218,7 +218,7 @@ class AstropyBuildDocs(SphinxBuildDoc):
             proc = subprocess.Popen([sys.executable, '-c', subproccode],
                                     stdin=subprocess.PIPE,
                                     stdout=subprocess.PIPE,
-                                    stderr=subprocess.STDOUT)
+                                    stderr=None)
 
             retcode = 1
             with proc.stdout:


Cadair:
If nobody has any objections to this, I will rebase it onto master. I was just doing it from inside the sunpy checkout which is still on 2.0.x.

Cadair:
Could someone give this a review?

astrofrog:
@eteq @Cadair Is this going to cause issues with https://github.com/astropy/astropy-helpers/pull/383?

Cadair:
Looks like they are solutions to the same problem. Personally, I think this one is cleaner, it is certainly "more unix".

Cadair:
for context with #383 you would do:
```
python setup.py build_docs --warning-file warnings.out
```

with this you would do:
```
python setup.py build_docs 2> warnings.out
```

Cadair:
nope, my last comment was BS.

Both are complementary. I have comments on how to do #383 in a different way though, I will put them there.

astrofrog:
I'm confused - this PR says 'don't capture stdout' but then the changes concern stderr?

Cadair:
whoops, it's about err.

PR #382, opened 2018-03-13
Make add_openmp_flags_if_available() work for clang

jamienoss:
The small openmp C test code, used to determine if openmp works, fails to build with clang due to missing include and library paths for openmp. When run, it then fails due to rpath (runtime path) issues.
    
This PR adds the necessary include, library, and runtime paths to the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,24 @@ matrix:
       env: PYTHON_VERSION=3.6 PIP_DEPENDENCIES='git+https://github.com/sphinx-doc/sphinx.git#egg=sphinx coveralls'
            CONDA_DEPENDENCIES="setuptools cython numpy pytest-cov"
 
+    # Test conda's clang
+    - os: osx
+      env:
+        - PYTHON_VERSION=3.5
+        - CONDA_DEPENDENCIES="setuptools sphinx cython numpy pytest-cov clang llvm-openmp"
+      before_install:
+        - export TEST_OPENMP="True"
+        - export CC=clang
+
+    # Test gcc on OSX
+    - os: osx
+      env:
+        - PYTHON_VERSION=3.5
+        - CONDA_DEPENDENCIES="setuptools sphinx cython numpy pytest-cov gcc"
+      before_install:
+        - export TEST_OPENMP="True"
+        - export CC=gcc
+
   # Uncomment the following if there are issues in setuptools that we
   # can't work around quickly - otherwise leave uncommented so that
   # we notice when things go wrong.
@@ -44,6 +62,16 @@ matrix:
   #   - env: PYTHON_VERSION=3.6 SETUPTOOLS_VERSION=dev DEBUG=True
   #          CONDA_DEPENDENCIES='sphinx cython numpy pytest-cov'
   #          EVENT_TYPE='push pull_request cron'
+
+before_install:
+
+      # Test OSX without OpenMP support
+      # Since the matrix OSX tests use the OS shipped version of clang, they double up
+      # as exploratory tests for when the shipped version has automatic OpenMP support.
+      # These tests will then fail and at such a time a new one should be added
+      # to explicitly remove OpenMP support.
+    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export TEST_OPENMP="False";
+      else export TEST_OPENMP="True"; fi
 
 install:
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,15 @@ astropy-helpers Changelog
   by the ``sphinx-astropy`` package in conjunction with the ``astropy-theme-sphinx``,
   ``sphinx-automodapi``, and ``numpydoc`` packages. [#368]
 
+- openmp_helpers.py: Make add_openmp_flags_if_available() work for clang.
+  The small openmp C test code, used to determine if openmp works, fails to
+  build with clang due to missing include and library paths for openmp.
+  When run, it then fails due to rpath (runtime path) issues.
+  Adds the necessary include, library, and runtime paths to the build.
+  Autogenerator utility added to gen openmp_enabled.py::is_openmp_enabled()
+  which can be called post build to determine state of OpenMP support.
+  [#382]
+
 
 3.0.2 (unreleased)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,11 +13,9 @@ astropy-helpers Changelog
   ``sphinx-automodapi``, and ``numpydoc`` packages. [#368]
 
 - openmp_helpers.py: Make add_openmp_flags_if_available() work for clang.
-  The small openmp C test code, used to determine if openmp works, fails to
-  build with clang due to missing include and library paths for openmp.
-  When run, it then fails due to rpath (runtime path) issues.
-  Adds the necessary include, library, and runtime paths to the build.
-  Autogenerator utility added to gen openmp_enabled.py::is_openmp_enabled()
+  The necessary include, library, and runtime paths now get added to the C test code
+  used to determine if openmp works.
+  Autogenerator utility added ``openmp_enabled.is_openmp_enabled()``
   which can be called post build to determine state of OpenMP support.
   [#382]
 
--- a/astropy_helpers/openmp_helpers.py
+++ b/astropy_helpers/openmp_helpers.py
@@ -13,13 +13,13 @@
 
 from __future__ import absolute_import, print_function
 
-import datetime
-import glob
 import os
-import subprocess
 import sys
-import tempfile
+import glob
 import time
+import datetime
+import tempfile
+import subprocess
 
 from distutils import log
 from distutils.ccompiler import new_compiler
@@ -43,8 +43,8 @@ int main(void) {
 
 def _get_flag_value_from_var(flag, var, delim=' '):
     """
-    Utility to extract `flag` value from `os.environ[`var`]` or, if not present,
-    from `distutils.sysconfig.get_config_var(`var`)`.
+    Utility to extract ``flag`` value from `os.environ[``var``]` or, if not present,
+    from `distutils.sysconfig.get_config_var(``var``)`.
     E.g. to get include path: _get_flag_value_from_var('-I', 'CFLAGS')
     might return "/usr/local/include".
 
@@ -59,8 +59,8 @@ def _get_flag_value_from_var(flag, var, delim=' '):
     # Simple input validation
     if not var or not flag:
         return None
-    l = len(flag)
-    if not l:
+    flag_length = len(flag)
+    if not flag_length:
         return None
 
     # Look for var in os.eviron
@@ -79,8 +79,8 @@ def _get_flag_value_from_var(flag, var, delim=' '):
     # Extract flag from {var:value}
     if flags:
         for item in flags.split(delim):
-            if flag in item[:l]:
-                return item[l:]
+            if flag in item[:flag_length]:
+                return item[flag_length:]
 
     return None
 
@@ -102,7 +102,7 @@ def get_openmp_flags():
     Notes
     -----
     The flags returned are not tested for validity, use
-    `test_openmp_support(openmp_flags=get_openmp_flags())` to do so.
+    `check_openmp_support(openmp_flags=get_openmp_flags())` to do so.
     """
 
     compile_flags = []
@@ -125,7 +125,7 @@ def get_openmp_flags():
 
     return {'compiler_flags':compile_flags, 'linker_flags':link_flags}
 
-def test_openmp_support(openmp_flags=None, silent=False):
+def check_openmp_support(openmp_flags=None):
     """
     Compile and run OpenMP test code to determine viable support.
 
@@ -135,8 +135,6 @@ def test_openmp_support(openmp_flags=None, silent=False):
         Expecting `{'compiler_flags':<flags>, 'linker_flags':<flags>}`.
         These are passed as `extra_postargs` to `compile()` and
         `link_executable()` respectively.
-    silent : bool, optional
-        silence log warnings
 
     Returns
     -------
@@ -180,14 +178,12 @@ def test_openmp_support(openmp_flags=None, silent=False):
             if len(output) == nthreads:
                 is_openmp_supported = True
             else:
-                if not silent:
-                    log.warn("Unexpected number of lines from output of test OpenMP "
-                             "program (output was {0})".format(output))
+                log.warn("Unexpected number of lines from output of test OpenMP "
+                         "program (output was {0})".format(output))
                 is_openmp_supported = False
         else:
-            if not silent:
-                log.warn("Unexpected output from test OpenMP "
-                         "program (output was {0})".format(output))
+            log.warn("Unexpected output from test OpenMP "
+                     "program (output was {0})".format(output))
             is_openmp_supported = False
     except (CompileError, LinkError):
         is_openmp_supported = False
@@ -202,7 +198,12 @@ def is_openmp_supported():
     Utility to determine whether the build compiler
     has OpenMP support.
     """
-    return test_openmp_support(silent=True)
+    
+    # don't log
+    log_threshold = log.set_threshold(log.FATAL)
+    ret = check_openmp_support()
+    log.set_threshold(log_threshold)
+    return ret
 
 def add_openmp_flags_if_available(extension):
     """
@@ -213,7 +214,7 @@ def add_openmp_flags_if_available(extension):
     """
 
     openmp_flags = get_openmp_flags()
-    using_openmp = test_openmp_support(openmp_flags=openmp_flags, silent=False)
+    using_openmp = check_openmp_support(openmp_flags=openmp_flags)
 
     if using_openmp:
         compile_flags = openmp_flags['compiler_flags'] if 'compiler_flags' in openmp_flags else None
@@ -239,7 +240,7 @@ def is_openmp_enabled():
 
 def generate_openmp_enabled_py(packagename, srcdir='.'):
     """
-    Utility for creating openmp_enabled.py::is_openmp_enabled()
+    Utility for creating `openmp_enabled.is_openmp_enabled()`
     used to determine, post build, whether the package was built
     with or without OpenMP support.
     """
--- a/astropy_helpers/openmp_helpers.py
+++ b/astropy_helpers/openmp_helpers.py
@@ -13,15 +13,17 @@
 
 from __future__ import absolute_import, print_function
 
-import os
-import sys
+import datetime
 import glob
-import tempfile
+import os
 import subprocess
+import sys
+import tempfile
+import time
 
 from distutils import log
 from distutils.ccompiler import new_compiler
-from distutils.sysconfig import customize_compiler
+from distutils.sysconfig import customize_compiler, get_config_var
 from distutils.errors import CompileError, LinkError
 
 from .setup_helpers import get_compiler_option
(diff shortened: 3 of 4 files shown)


jamienoss:
Let me know if there's a better/neater/preferred way of doing any of this.

I wasn't sure about the content changes to CHANGES.rst, please advise.

jamienoss:
This is required for https://github.com/astropy/astropy/pull/7293

astrofrog:
Thanks! I will review this next week.

astropy_helpers/openmp_helpers.py
```diff
@@ -39,6 +41,34 @@
 }
 """
 
+def _split_option_from_var(option, var, delim=' '):
```
astrofrog: This function needs a docstring for clarity (astropy-helpers code has typically been under-documented in the past, causing problems down the road, so we need to make sure all new code is easy to understand/read). Also please include some more comments below in the function to explain a bit more what is going on.

astropy_helpers/openmp_helpers.py
```diff
@@ -98,10 +144,44 @@ def add_openmp_flags_if_available(extension):
         os.chdir(start_dir)
 
     if using_openmp:
-        log.info("Compiling Cython extension with OpenMP support")
-        extension.extra_compile_args.append(compile_flag)
-        extension.extra_link_args.append(link_flag)
+        log.info("Compiling Cython/C/C++/ extension with OpenMP support")
```
astrofrog: remove / after C++

astropy_helpers/openmp_helpers.py
```diff
@@ -39,6 +41,34 @@
...
+                return flag[l:]
+    return None
+
+def _get_include_path():
+    return _split_option_from_var('-I', 'CFLAGS')
+
+def _get_library_path():
+    return _split_option_from_var('-L', 'LDFLAGS')
 
 def add_openmp_flags_if_available(extension):
```
astrofrog: Add a note in the docstring that one can pass ``None`` to find out if OpenMP compilation would work without doing it for a specific extension.

astropy_helpers/openmp_helpers.py
```diff
@@ -98,10 +144,44 @@ def add_openmp_flags_if_available(extension):
...
+    """
+    Utility for creating openmp_enabled.py::is_openmp_enabled()
+    used to determine, post build, whether the package was built
+    with or without OpenMP support.
+    """
+
+    if packagename.lower() == 'astropy':
+        packagetitle = 'Astropy'
+    else:
+        packagetitle = 'Astropy-affiliated package ' + packagename
```
astrofrog: Not all packages using astropy-helpers are affiliated packages, so I would just set packagetitle = packagename

astropy_helpers/openmp_helpers.py
jamienoss: Will do. Maybe it's also a good idea to make it the default, `extension=None`, and also add a "wafer thin" wrapper with a more semantically correct name?

jamienoss: Should probably suppress the exception when `None` also(?)

astrofrog: I don't know if it's super important to have the semantically correct name, you can always just put a comment above the call to the function to explain what it does in that case?

astropy_helpers/openmp_helpers.py
jamienoss: I'm going to slightly refac this and rename it, sorry.

jamienoss:
@astrofrog 
>How portable is the solution related to LDFLAGS and CFLAGS on e.g Windows? Should we check that the platform is Linux/MacOS X and only fetch the extra flags from there?

Good call, I'll make this conditional for now and we can address later(?).

astrofrog:
> Good call, I'll make this conditional for now and we can address later(?).

Sounds good!

jamienoss:
@astrofrog 
Regarding the Travis test, do you want to keep this exploratory as the following [comment](https://github.com/astropy/astropy-helpers/blob/master/astropy_helpers/tests/test_openmp_helpers.py#L32-L36) suggests or explicit? By this I mean, should I make the assertion conditional upon the version of clang/llvm-openmp?
```
 # Make sure that on Travis (Linux) and AppVeyor OpenMP does get used (for
    # MacOS X usually it will not work but this will depend on the compiler).
    # Having this is useful because we'll find out if OpenMP no longer works
    # for any reason on platforms on which it does work at the time of writing.
    # OpenMP doesn't work on Python 3.x where x<5 on AppVeyor though.
```

jamienoss:
@astrofrog 
I made requested changes, though, upon looking at it again, I didn't like the hacky overloaded use of  `add_openmp_flags_if_available(extension)` so I split it up. Sorry, I know this means more for you to review.

Regarding the new test `test_generate_openmp_enabled_py()`, I was really only taking a guess here, 
I have no idea of the test framework when it comes to writing to the filesystem during the test. I took a look at some in `test_git_helpers.py` for the autogeneration of `version.py` but it looked liked overkill for what I needed, but perhaps this is the part where I have no idea.

jamienoss:
Doh!

jamienoss:
@astrofrog hold.....

jamienoss:
@astrofrog and go...

Sorry, I messed up the testing.

I'm looking into what can be done about the OSX test but you're all good to review otherwise. Cheers

.travis.yml
```diff
@@ -21,6 +21,9 @@ env:
     - CONDA_DEPENDENCIES="setuptools sphinx cython numpy pytest-cov"
     - PIP_DEPENDENCIES="coveralls"
     - EVENT_TYPE='push pull_request'
+    - CFLAGS="-m64"
+    - LDFLAGS="-m64"
```
astrofrog: Are these flags needed on all platforms? If not, I would favor adding them only to the relevant builds below.

.travis.yml
```diff
@@ -36,6 +39,49 @@ matrix:
...
+        - export PATH=$path"/bin":$PATH
+        - export C_INCLUDE_PATH=$path"/include":$C_INCLUDE_PATH
+        - export LD_LIBRARY_PATH=$path"/lib":$LD_LIBRARY_PATH
+        - export CC=$path"/bin/clang"
+        - svn co http://llvm.org/svn/llvm-project/openmp/trunk openmp
+        - mkdir openmp-build && cd openmp-build
+        - cmake ../openmp
+        - make
+        - make install
+        - cd ../
```
astrofrog: In order to keep the Travis config file as clean as possible, could you move all this into a script such as ``.install_llvm_clang.sh`` and just call that here?

.travis.yml
```diff
@@ -36,6 +39,49 @@ matrix:
...
+    # Since this is using the OS shipped version of clang, this test doubles up
+    # as an exploritory test for when the shipped version has automatic OpenMP support.
+    # This test will then fail and at such a time should not be reomved but
+    # altered to explicitly remove OpenMP support.
+    - os: osx
+      env:
+        - PYTHON_VERSION=3.5
+        - export TEST_OPENMP="False"
+      before_install:
+        - echo "Default before_install ignored"
```
astrofrog: Rather than overriding before_install here to disable it, what about swapping this build with the one that has the default ``before_install``? Then in the default ``before_install``, just set TEST_OPENMP to False if the platform is OSX.

.travis.yml
astrofrog: And actually I would recommend the same for the two other cases where you install things manually too - one script for each. This is just so that the install/setup cases are self contained and we can easily run them locally.

jamienoss:
> I'm very sorry for taking a while to get back to this

No problem, I have been on vacation for the past two weeks anyhow.

> just to make sure I understand, this still doesn't work out of the box with Apple's default clang, correct?

Correct, though I don't really understand why as I think it should.

>One thing I don't understand is - is it then still the case that if using clang installed manually, the current OpenMP code still didn't work because -openmp doesn't work for clang, regardless of whether it's the built-in one or one installed by e.g. homebrew?

It will work when installed manually for LLVM/Clang v>=3.7 ([ref](https://clang-omp.github.io/)). The issue is that Apple's `clang` is based on the root LLVM and the version numbers don't track. According to [wikipedia](https://en.wikipedia.org/wiki/Xcode) Xcode 6.3 is based on LLVM 3.6 and should therefore except `-fopenmp`, I have no idea why it doesn't. I would like to get to the bottom of the issue, but don't currently have the time to do so.

.travis.yml
jamienoss: I'll double check and adjust accordingly.

.travis.yml
jamienoss: The idea is to test both `TEST_OPENMP` states but since there are two default OSX builds I can highjack one if you're cool with that?  I'd prefer not to though, tests shouldn't really overlap. The two counterparts are:
```python
env:
  matrix:
    - PYTHON_VERSION=3.5

- os: osx
      env:
        - PYTHON_VERSION=3.5
        - export TEST_OPENMP="False"
      before_install:
        - echo "Default before_install ignored"
```

.brew_install_gcc6.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
```
astrofrog: Could you include comments/descriptions at the top of each file to just explain in a verbose way which versions of the compilers you are setting up?

.travis.yml
astrofrog: I guess what I meant is that the original OSX test was the one without OpenMP anyway, so why not make the three new ones the ones that test different OpenMP enabled compilers, and keep the default build as the one with the default compilers?

.brew_install_gcc6.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
```
astrofrog: Could you include comments/descriptions at the top of each file to just explain in a verbose way which versions of the compilers you are setting up?

astropy_helpers/openmp_helpers.py
```diff
@@ -39,69 +41,222 @@
...
+    Notes
+    -----
+    The flags returned are not tested for validity, use
+    `test_openmp_support(openmp_flags=get_openmp_flags())` to do so.
+    """
 
-    start_dir = os.path.abspath('.')
+    compile_flags = []
+    include_path = _get_include_path()
+    if include_path:
```
astrofrog: Could you add a comment explaining why you need to add library and include paths - e.g. does Python ignore environment variables by default? Sorry for being picky with comments, but we've been burned in the past with undocumented workarounds for issues in astropy-helpers, so I'm just making sure this is understandable by other developers :)

jamienoss: I don't think commenting in `get_openmp_flags` is correct, semantically the use for why you want these flags belongs to the caller as they don't actually have to be passed to a compiler at this stage, they may just be checked or printed etc. `_get_flag_value_from_var` already docs extraction method, i.e. from `os.environ[`var`]` or, if not present, from `distutils.sysconfig.get_config_var(`var`)`.

I believe the level of doc you want already exists [here](https://github.com/jamienoss/astropy-helpers/blob/9ba2f37481bdd2e5376531f91d261c6c35b9a2c5/astropy_helpers/openmp_helpers.py#L151-L156), where they are actually being used as flags passed to the compiler.

```python 
        # customize_compiler() extracts info from os.environ. If certain keys
        # exist it uses these plus those from sysconfig.get_config_vars().
        # If the key is missing in os.environ it is not extracted from 
        # sysconfig.get_config_var(). E.g. 'LDFLAGS' get left out, preventing
        # clang from finding libomp.dylib because -L<path> is not passed to linker.
        # Call get_openmp_flags() to get flags missed by customize_compiler().
```

Is this cool as is?

jamienoss:
@astrofrog FYI I'm still actively ironing out some Travis OSX issues.

jamienoss:
...and now other timeout issues. Can someone restart this test for me, please?

jamienoss:
Ok, this got ugly quickly.

It is now dependent on https://github.com/astropy/ci-helpers/pull/291

The last changes should probably be either a separate (initial) commit or PR, up to you guys? Just let me know and I'll reorder etc.

https://github.com/travis-ci/travis-ci/issues/6307 was the culprit.

@astrofrog & @bsipocz let me know your thoughts.

bsipocz:
I haven't followed this PR along closely, but now it really seems to turn a bit into a hacky direction. 

I assume all of these compiler install workarounds can be useful for any packages. If so, I suggest to move it to ci-helpers and structure a bit more to be able to use one env variable to control it (`COMPILER_VERSION` or similar). Also we could add a `BREW_DEPENDENCIES` at some point, but that is probably beyond the scope of this PR and feature.

But even if we don't move the CI stuff,  I feel it would be a much better structure to dump the hidden files into a subdirectory rather than have them as hidden files at the top level.

jamienoss:
> I haven't followed this PR along closely, but now it really seems to turn a bit into a hacky direction.

Regarding the tests, I agree. There's not much scope for elegance within the Travis framework - it is relient on too much overloading. I guess this is why `ci-helpers` exists.

>But even if we don't move the CI stuff, I feel it would be a much better structure to dump the hidden files into a subdirectory rather than have them as hidden files at the top level.

I like this also and will change.

jamienoss:
Regarding `ci-helpes`, all of the scripts can be added to it then we just add a single one to`astropy-helpers` to do the dir check and clone.

Create an `install_scripts` dir to contain as it says on the tin and then add the bash_script_saver.sh to the current Travis dir.

astrofrog:
By the way, just a thought - could one option be to include a test against the compilers in conda, which I think should support OpenMP?

jamienoss:
> By the way, just a thought - could one option be to include a test against the new compilers in conda, which I think should support OpenMP?

This was the first thing I tried and it didn't work. I tried it again, just now, as a sanity check but still nothing. I don't know why as this works fine on my machine, it was how I initially installed an OpenMP working clang, which is why I tried it first. Any ideas what might be going on?

Also, there is an unstable failure of the `import` of the generated `openmp_enabled`, but only on OSX and only sporadically will it import ok.
```
E       ImportError: No module named 'openmp_enabled'
```
Any ideas here? I tried adding in a sleep before the import (rationale=magic) but it did nothing. I wondered if it might be a caching issue but couldn't think how.

I would gratefully appreciate some affirmative direction on this PR as I need to move on and spend my time elsewhere.

jamienoss:
I used `- export CC=/anaconda/bin/clang` maybe this is not correct for the Travis build?

jamienoss:
Trying `/Users/travis/miniconda/bin/clang`...

jamienoss:
@astrofrog I think this should be all done now.

I got clang working via conda, I did the same for gcc on OSX to drop all of the scripts and reduce the number of OSX tests.

I addressed the import stability issue by appending the current dir, `'.'`, to `sys.path`, this shouldn't be needed but it seems to work. Here are the last three successful test builds
* [build 1550](https://travis-ci.org/astropy/astropy-helpers/builds/369206239?utm_source=github_status&utm_medium=notification)
* [build 1549](https://travis-ci.org/astropy/astropy-helpers/builds/369180199?utm_source=github_status&utm_medium=notification)
 * [build 1548](https://travis-ci.org/astropy/astropy-helpers/builds/369167583?utm_source=github_status&utm_medium=notification)

.travis.yml
```diff
@@ -45,6 +63,16 @@ matrix:
   #          CONDA_DEPENDENCIES='sphinx cython numpy pytest-cov'
   #          EVENT_TYPE='push pull_request cron'
 
+before_install:
+
+      # Test OSX without OpenMP support
+      # Since the matrix OSX tests use the OS shipped version of clang, they double up
+      # as exploritory tests for when the shipped version has automatic OpenMP support.
```
astrofrog: exploritory -> exploratory

.travis.yml
```diff
@@ -45,6 +63,16 @@ matrix:
...
+      # These tests will then fail and at such a time a new one should be added
+      # to explicitly remove OpenMP support.
+    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export TEST_OPENMP="False";
+      else export TEST_OPENMP="True"; fi
```
astrofrog: Minor: I'd set TEST_OPENMP=True as a global higher up and just set it to False on MacOS X

jamienoss: It's never good design to split conditionals like this. State logic should be as independent as it can be and contained within a single place. Splitting it out increases complexity and dependency, this is especially true here when considering stage overloading. Basically, it should only be set once and since this can't exist as a global (due to it be conditional on OS), it needs to exist within an actual eval stage.

If `before_install` is overloaded and `TEST_OPENMP` not set, the test will raise an exception [here](https://github.com/astropy/astropy-helpers/pull/382/files#diff-58c5d1bcd09b08d0f8c8144ada928f7eR22) alerting the developer that they have forgotten to add this to their overloaded version.

jamienoss:
> So just to check, do we still need the added complexity in terms of checking environment variables in the OpenMP helpers? Or do the conda compilers support -openmp properly?

Yes, they are needed to find and link with libomp. `-fopenmp` just tells the compiler to except the OpenMP pragma/directives and, when used as a wrapper to the linker, to link with libomp. With gcc, the library is packaged and shipped with the core libraries so only `-fopenmp` is needed, this is generally not true of the others though.

> ...I just want to make sure all the other changes are still needed.

They are.

jamienoss:
@astrofrog 
>> ...I just want to make sure all the other changes are still needed.
>
>They are.

Well... the one thing I am slightly unsure of is how the path flags play out, especially `-rpath`, when it gets built and released by conda. I don't know how to test this either.

I just spoke to Matt Rendina about this and I can try it out using `conda-build` and then installing it into a diff conda env. I will do so now.

jamienoss:
_c.f._ https://github.com/astropy/astropy-helpers/pull/389#issuecomment-384755136

jamienoss:
_Note:_ experiencing https://github.com/astropy/astropy/issues/6424

astrofrog:
@jamienoss - just to check, is this ready to merge from your side?

jamienoss:
@astrofrog 
>just to check, is this ready to merge from your side?

Yes and no, sorry. I finally got the conda build working, using https://github.com/astropy/astropy/pull/7293 as the test case. It built, tests passed but only when using gcc due to conda's LLVM (v4.0.1) bug, which is why I originally went with the manualy installs here. 

What I haven't managed to get around to, is taking that build and trying to install it on another machine or when anaconda has a diff path (i.e. not root). This last test would test any rpath concerns.

Other than this though, it is good by me.

jamienoss:
@astrofrog Just tested it, it works. Only thing is that, again, this is with gcc and not clang and llvm, which this PR is semantically designed to address. The rpaths didn't create an issue. I checked them using `otool -L` (on lib_convolve*.so) and conda had replaced the path literals. I think this should be enough, without testing on the actual conda build machines/env.

What are your thoughts?

.travis.yml
```diff
@@ -45,6 +63,16 @@ matrix:
...
 
+before_install:
+
+      # Test OSX without OpenMP support
+      # Since the matrix OSX tests use the OS shipped version of clang, they double up
+      # as exploratory tests for when the shipped version has automatic OpenMP support.
+      # These tests will then fail and at such a time a new one should be added
+      # to explicitly remove OpenMP support.
+    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export TEST_OPENMP="False";
+      else export TEST_OPENMP="True"; fi
```
bsipocz: Remove the `before_install` section and define the env variables in the jobs above.

.travis.yml
```diff
@@ -36,6 +36,24 @@ matrix:
       env: PYTHON_VERSION=3.6 PIP_DEPENDENCIES='git+https://github.com/sphinx-doc/sphinx.git#egg=sphinx coveralls'
            CONDA_DEPENDENCIES="setuptools cython numpy pytest-cov"
 
+    # Test conda's clang
+    - os: osx
+      env:
+        - PYTHON_VERSION=3.5
+        - CONDA_DEPENDENCIES="setuptools sphinx cython numpy pytest-cov clang llvm-openmp"
+      before_install:
```
bsipocz: I'm pretty sure there is no need for these `before_install` sections either, put those variable into the `env` section.

CHANGES.rst
```diff
@@ -12,6 +12,15 @@ astropy-helpers Changelog
   by the ``sphinx-astropy`` package in conjunction with the ``astropy-theme-sphinx``,
   ``sphinx-automodapi``, and ``numpydoc`` packages. [#368]
 
+- openmp_helpers.py: Make add_openmp_flags_if_available() work for clang.
```
bsipocz: This changelog is way too verbose, any chance to cut it back a bit? E.g. no need for filenames, but can use double backticks to highlight the name of the new functions.

astropy_helpers/openmp_helpers.py
```diff
@@ -13,15 +13,17 @@
...
 from __future__ import absolute_import, print_function
 
+import datetime
+import glob
 import os
+import subprocess
 import sys
-import glob
 import tempfile
-import subprocess
```
bsipocz: Please don't shuffle the imports around but just add the new ones, they were sorted by length on purpose.

astropy_helpers/openmp_helpers.py
```diff
@@ -39,69 +41,222 @@
 }
 """
 
+def _get_flag_value_from_var(flag, var, delim=' '):
+    """
+    Utility to extract `flag` value from `os.environ[`var`]` or, if not present,
```
bsipocz: use double backticks for parameters as a good practice, in public functions it would trip sphinx over otherwise (single backticks are for intersphinx links)

astropy_helpers/openmp_helpers.py
```diff
@@ -39,69 +41,222 @@
...
+    from `distutils.sysconfig.get_config_var(`var`)`.
+    E.g. to get include path: _get_flag_value_from_var('-I', 'CFLAGS')
+    might return "/usr/local/include".
```
bsipocz: for which OS? Please keep comments it OS agnostic

astropy_helpers/openmp_helpers.py
```diff
@@ -39,69 +41,222 @@
...
-        link_flag = '-fopenmp'
+        compile_flags.append('-fopenmp')
+        link_flags.append('-fopenmp')
+        if lib_path:
+            link_flags.append('-Wl,-rpath,' + lib_path)
 
-    try:
+    return {'compiler_flags':compile_flags, 'linker_flags':link_flags}
+
+def test_openmp_support(openmp_flags=None, silent=False):
```
bsipocz: rename this function to `check_openmp_support` or something, it's very confusing to see a test function in the main codebase (and I'm not 100% certain that pytest likes it either).

astropy_helpers/openmp_helpers.py
```diff
@@ -39,69 +41,222 @@
...
+    """
+    Compile and run OpenMP test code to determine viable support.
+
+    Parameters
+    ----------
+    openmp_flags : dictionary, optional
+        Expecting `{'compiler_flags':<flags>, 'linker_flags':<flags>}`.
+        These are passed as `extra_postargs` to `compile()` and
+        `link_executable()` respectively.
+    silent : bool, optional
```
bsipocz: Why is this parameter needed? Surely the verbosity can already been tuned via the logger

astropy_helpers/openmp_helpers.py
```diff
@@ -39,69 +41,222 @@
...
-    printed to the console and no flags will be added)
 
-    Returns `True` if the flags were added, `False` otherwise.
+    if sys.platform.startswith('win'):
+        return None
+
+    # Simple input validation
+    if not var or not flag:
+        return None
+    l = len(flag)
```
bsipocz: if possible avoid single letter variables

astropy_helpers/openmp_helpers.py
jamienoss: Woops, sorry, probably an auto-format.

astropy_helpers/openmp_helpers.py
jamienoss: Ok, cool, thanks for the intel, will do.

astropy_helpers/openmp_helpers.py
jamienoss: Good point, I hadn't thought of the pytest 'test_' namespace clash, cheers.

astropy_helpers/openmp_helpers.py
jamienoss: This is why I added the "E.g." and "might".

.travis.yml
jamienoss: I'll double check this, but if I remember correctly, it needs to be here so as to also work with the matrix permutations. There is no way to add this to them as they will be evaluated not at the correct stage, i.e.  TRAVIS_OS_NAME is undefined. For the explicit tests it is, which is why I do so there also, but it isn't for the `env:matrix:` sec.

.travis.yml
jamienoss: So, I added these here following the Travis documentation. Also, because of the generic `before_install`, which then has to be overloaded per explicit job.

jamienoss: _c.f._ https://github.com/astropy/astropy-helpers/pull/382/files#r191006134

astropy_helpers/openmp_helpers.py
jamienoss: Do you mean import name length? If so, I should insert the two additions accordingly, right, rather than just append them to the list?

astropy_helpers/openmp_helpers.py
jamienoss: How? I couldn't find anything other than setting ``log.set_threshold(FATAL)``, which I am unsure of its semantic correctness.

.travis.yml
jamienoss: Regardless of OS, ``TEST_OPENMP`` has to be set to something. If set as a global env and to ``true`` it will prevent the OSX tests from passing. If set to ``false`` it will prevent the linux tests from passing (unless we explicitly turn off the openmp support - which is altogether silly). One alternative is to make it such that all global matrixed OS tests pass with either state, which probably means getting OpenMP to work for _all_ the OSX tests. The other alternative is to make  ``TEST_OPENMP`` conditional and this therefore can't happen in the global env stage by definition. It has to go in the ``install`` or better still (due to overloading and its current use, not to mention semantically) in the ``before_install`` stage. Let me know if there is an alternative, perhaps the entire idea needs to be re-thought, however, I honestly don't see the cause to _not_ use the ``before_install_ stage.

astrofrog:
I tested this out locally for astropy-healpix and it does indeed work nicely. It also looks like @bsipocz's comments have been addressed so I'll go ahead and merge this. Thanks @jamienoss!

PR #383, opened 2018-03-18
Add warning file argument to build_docs

eteq:
This PR adds an option to output a "warnings file" from the `build_docs` command. (Inspired by the similar option that now exists in the ``sphinx-build`` command line tool.)  

Much more importantly, when used on travis, it will also automatically output the results of that command at the end of the travis build log.  This will make it so that docs build failures no longer require hunting through a thousand lines of output: the warnings and *only* the warnings will be at the bottom of the build log.

To activate this in astropy will require a second PR that just modifies the travis matrix to add an argument like `` --warning-file warnings.out`` to the `build_docs` command.  Will do that myself if this PR is approved.

@astrofrog @bsipocz - I tentatively milestoned this 3.1.0, since it's a new feature... but we might want this included in the build *before* the 3.1 core package release, as I think it will help quite a bit in diagnosing travis docs failures.  I suppose the PR that updates the travis file can update astropy to this helpers version as an interim solution?

### DIFF
--- a/astropy_helpers/commands/build_sphinx.py
+++ b/astropy_helpers/commands/build_sphinx.py
@@ -263,12 +263,19 @@ class AstropyBuildDocs(SphinxBuildDoc):
                     # this means we are in the travis build, so customize
                     # the message appropriately.
                     msg = ('The build_docs travis build FAILED '
-                           'because sphinx issued documentation '
-                           'warnings (scroll up to see the warnings).')
+                           'because sphinx issued documentation warnings. '
+                           'Scroll up to see the warnings in context.')
+                    if self.warning_file:
+                        msg += 'Warnings/errors are also listed below:\n'
+                        with open(os.path.abspath(self.warning_file)) as f:
+                            msg += f.read()
                 else:  # standard failure message
                     msg = ('build_docs returning a non-zero exit '
                            'code because sphinx issued documentation '
                            'warnings.')
+                    if self.warning_file:
+                        fn = os.path.abspath(self.warning_file)
+                        msg += ' Warnings recorded in file: "{}"'.format(fn)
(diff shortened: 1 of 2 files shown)


eteq:
Oh and one possible perturbation on this would be to have the warnings come *before* the travis message that currently says "The build_docs travis build FAILED because sphinx issued documentation warnings (scroll up to see the warnings)."  With this PR that will no longer be the last message, as any warnings come after that.  But if we want to preserve that at the end, I could output the content of the warning file *before* that message.

bsipocz:
@eteq - We also promised to keep shipping new features in the 2.0.x branch for python2 users, so if it's OK with you we may want to try to backport this.

astropy_helpers/commands/build_sphinx.py
```diff
@@ -179,6 +203,13 @@ def run(self):
                 subproccode[i] = repr(val)
         subproccode = ''.join(subproccode)
 
+        if self.warning_file:
+            # VERY hacky: assumes that the text below exists and only in the right
+            # place.  Could change in future sphinxes
```
astrofrog: Is there a better solution, or is this the best we can do?

Cadair:
@eteq rather than doing funky stuff with `Tee`, can you not just reconfigure the sphinx logging module to add a new handler to warnings and output to a file?

i.e. a modified version of this:
https://github.com/sunpy/sunpy/pull/2553/files#diff-85987f48f1258d9ee486e3191495582dR33

bsipocz:
@eteq - can you rebase please?

astrofrog:
Note that this is no longer really needed since the warnings are now shown at the end of the log in the current implementation.

PR #384, opened 2018-03-19
Remove ez_setup.py

astrofrog:
As described in https://github.com/astropy/astropy-helpers/issues/297, ``ez_setup.py`` is deprecated. In fact, it's very rarely needed - these days virtually everyone has setuptools installed:

* If a user does `pip install ...`, pip requires setuptools so setuptools will be present

* If a user has conda and is doing e.g. `python setup.py ...`, conda environments automatically include setuptools.

* Most other major scientific packages such as Numpy, Scipy, Pandas, etc. just assume setuptools is present.

I think this is safe to remove. Once it is removed here we can update astropy-helpers in the core package and remove it there too, as well as in the package template.

Now that setuptools 1.0 has been out for a while I think it's also safe to simply remove the logic related to 'distribute' and require setuptools 1.0 or later. For user-friendliness, I've made it so that a nice error message is printed rather than a traceback.


CHANGES.rst
```diff
@@ -6,6 +6,8 @@ astropy-helpers Changelog
 
 - Removing deprecated test_helpers.py file. [#369]
 
+- Removing ez_setup.py file and require setuptools 1.0 or later. [#384]
```
eteq: (nitpick) to be grammatically correct I think this needs to be "requiring"?  Or Removing->remove.

bsipocz:
@astrofrog - So you suggest to backport this to LTS (https://github.com/astropy/package-template/issues/315#issuecomment-388365808), which is fine by me. But then we also need to backport to 3.0.x otherwise the changelog will become a mess. Would that be OK with you?

astrofrog:
Yes that sounds good, backporting to v2.0.x and v3.0.x seems like the right thing to do.

PR #385, opened 2018-03-21
Add version_info tuple to autogenerated version.py

jamienoss:
Allows for simple version checking, i.e. version_info > (2,0,1), in the same
manner that sys.version_info would be checked.


pllim:
FWIW there is `minversion()` in `astropy` (see https://github.com/astropy/astropy/blob/master/astropy/utils/introspection.py#L90). FYI.

jamienoss:
Rebased and conflicts resolved.

PR #389, opened 2018-04-20
Move clone of ci-helpers to script

jamienoss:
If astropy/ci-helpers is needed, it would useful to have the clone
contained as reusable functionality. This function first checks to
see if it has already been cloned before doing so.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ matrix:
 
 install:
 
-    - git clone git://github.com/astropy/ci-helpers.git
+    - source scripts/get_ci-helpers.sh
     - source ci-helpers/travis/setup_conda.sh
 
     # We cannot install the developer version of setuptools using pip because
--- a/scripts/README.rst
+++ b/scripts/README.rst
@@ -1,0 +1,8 @@
+Scripts
+=======
+
+This dir is designed to contain only scripts.
+
+Ideally it should contain only those scripts generic and necessary to
+``astropy-helpers``, i.e. ``get_ci-helpers.sh``. Any script less generic, e.g.
+those used for CI, should live elsewhere, e.g. https://github.com/astropy/ci-helpers.
--- a/scripts/get_ci-helpers.sh
+++ b/scripts/get_ci-helpers.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+# Check whethere ci-helpers has already been cloned, if not, clone.
+# Assumes working dir is correct
+
+if [ ! -d "ci-helpers" ]; then git clone --depth 1 git://github.com/astropy/ci-helpers.git; fi


jamienoss:
@bsipocz If you want this, let me know which target ver and I'll add to `CHANGES.rst`.

bsipocz:
@jamienoss - Please don't move it to a separate script. The current structure is shared between all packages in the ecosystem, and while `astropy-helpers` doesn't use the package template, the layout of the CI config files are the same as everywhere else making maintenance and automatic updates, etc. easier.

jamienoss:
@bsipocz 
> Please don't move it to a separate script. The current structure is shared between all packages in the ecosystem, and while astropy-helpers doesn't use the package template, the layout of the CI config files are the same as everywhere else making maintenance and automatic updates, etc. easier.

Ok, but just so that I understand more, this change doesn't break anything, does it? I mean, if it's the calling of the new func/script from `.travis.yml`, then this can be returned to the naked clone that it was before.

The rationale behind wrapping the clone is so that it does a pre-check that would otherwise need to be added manually anytime that anything from `ci-helpers` was needed. This is certainly true for its use with Travis due to the possible overloading of build stages and/or needing it before the `install` stage (where it is currently cloned).

For example, to execute a script from a Travis OSX build, you would need to use the `rvm` workaround ([here](https://github.com/astropy/ci-helpers/blob/master/travis/setup_conda_osx.sh#L3-L10)) so that it doesn't prematurely fail. That `rvm` code should be self-contained (_c.f._ https://github.com/astropy/ci-helpers/pull/291) so that it can be called directly. To do so from `astrop-helpers` requires first obtaining `ci-helpers`. This would need to be done everytime a script was executed as there is no strong assurance that it has already been cloned.

astrofrog:
@jamienoss - I'm :-1: on this change.

> Ok, but just so that I understand more, this change doesn't break anything, does it?

It breaks the assumption that this line is in the Travis file, which we may need to rely on in future if deploying updates to the command.

ci-helpers is intended to be used only in a CI environment, so we know for sure that ci-helpers just needs to be cloned once and only once at the start of the CI process, so we don't need to worry about checking whether it's needed or not. Does this make sense?

jamienoss:
>ci-helpers is intended to be used only in a CI environment, so we know for sure that ci-helpers just needs to be cloned once and only once at the start of the CI process, so we don't need to worry about checking whether it's needed or not. Does this make sense?

The issue here is that it is _not_ being cloned at the _beginning_ of the CI process. It is currently cloned within the `install` stage. The canonical dependency installation stage, as noted by the [Travis docs](https://docs.travis-ci.com/user/installing-dependencies/), is the `before_install` stage.  If anything from `ci-helpers` is required within the `before_install` stage then it will have to be cloned then and there. This then requires a subsequent check, to see if it's already been cloned, from the `install` stage before it tries to clone. Since all of these stages can be overloaded per test, there is no way to globally clone `ci-helpers` once, at the beginning, and be certain that it always has been. Therefore, the pre-check between `before_install` and `install` is true for any stage transition, or per test due to the possibility of stage overloading. Basically, there is no guarenteed _beginning_ of the CI process, it is test dependent.

There are only three(ish) other ways to address this, unwrap the code snippet and use naked everytime it is needed (ugly), constrain the use and overloading of build stages (not a good idea), or move the clone to the `before_install` stage (and add a comment about overloading).

> It breaks the assumption that this line is in the Travis file, which we may need to rely on in future if deploying updates to the command.

The only assumption I see here is the use of `--depth`. If you're arguing that the clone command may need to be changed in the future to be used by something other than a Travis test, this is easily rectified by splitting and tayloring the scripts. For example, this can already be achieved by, instead, adding 2 scripts: `get_ci-helpers.sh` and `get_ci-helpers_for_travis.sh`. The later containing the `--depth 1` and the former not, using `--depth` only as an example here.

I feel that at least something is needed to prevent someone else from inevitably stumbeling in the future. What about just moving it to the `before_install` stage?

bsipocz:
Yes, moving the cloning to the `before_install` is a good workaround for now. But, still, there seem to be too many unnecessary changes, exports. In fact, I fail to see why `before_install` is needed now that stock compilers are used only.

And if we go ahead and decide to test with self build compilers, then that will need to be dealt with in ci-helpers (especially if we start using it in core, etc.) so there won't be any need for `before_install` either. Sadly, I have zero time to deal with any of this any time soon.

jamienoss:
> Moving to the before_install stage would be fine by me - @bsipocz, what do you think?

Cool

> But, still, there seem to be too many unnecessary changes, exports. In fact, I fail to see why before_install is needed now that stock compilers are used only.

I believe you're referring to https://github.com/astropy/astropy-helpers/pull/382 rather than this PR.  Regarding that PR and your comment - `clang` and `gcc` are specified as `CONDA_DEPENDENCIES`, true, however, it is still required that you explicitly state their use as the C compiler via `CC`. The basic rule is that tests should be explicit. So backtracking on the `export`s slightly, the idea is not to use the default `Xcode` compilers. When conda installing `gcc`, the `export` isn't needed as this will overwrite the `Xcode` clang imposter (assuming CC=gcc). When conda installing `clang` it will overwrite Apple's clang so that `export`s not needed (assuming CC=clang). Back-peddling on my backtracking, they are still needed, as you still have two compilers that you explcitly wish to test, i.e. the "assuming"s don't hold. (I'd like to add a clang linux test but don't currently have time to deal with any possible fallout). 

> And if we go ahead and decide to test with self build compilers, then that will need to be dealt with in ci-helpers (especially if we start using it in core, etc.) so there won't be any need for before_install either. Sadly, I have zero time to deal with any of this any time soon.

Isn't the correct use of a Travis build to use the `install` build stage for the installation of the repo package itself and the `before_install` for pre-repo-build installations? Even with custom scripts in `ci-helpers`, they should still be `sourced` by the `before_install` stage, right?

jamienoss:
Closing: moved to #390

bsipocz:
There is no any practical difference between `before_install` and `install` except that they executed after each other. The repo being tested should be built in the `script` section, so a failure there is handled differently than a failure in building the dependencies.

jamienoss:
> There is no any practical difference between before_install and install except that they executed after each other. The repo being tested should be built in the script section, so a failure there is handled differently than a failure in building the dependencies.

Ok, thanks, I get you now.

bsipocz:
No worries. It sometimes have a strange logic, but Travis is still the best configurable and flexible CI provider out there.

PR #395, opened 2018-06-18
Minor updates to OpenMP infrastructure

astrofrog:
Just some small updates to the OpenMP infrastructure

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ matrix:
         - PYTHON_VERSION=3.5
         - CONDA_DEPENDENCIES="setuptools sphinx cython numpy pytest-cov clang llvm-openmp"
       before_install:
-        - export TEST_OPENMP="True"
+        - export OPENMP_EXPECTED="True"
         - export CC=clang
 
     # Test gcc on OSX
@@ -51,7 +51,7 @@ matrix:
         - PYTHON_VERSION=3.5
         - CONDA_DEPENDENCIES="setuptools sphinx cython numpy pytest-cov gcc"
       before_install:
-        - export TEST_OPENMP="True"
+        - export OPENMP_EXPECTED="True"
         - export CC=gcc
 
   # Uncomment the following if there are issues in setuptools that we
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -42,7 +42,7 @@ matrix:
         - PYTHON_VERSION=3.5
         - CONDA_DEPENDENCIES="setuptools sphinx cython numpy pytest-cov clang llvm-openmp"
       before_install:
-        - export TEST_OPENMP="True"
+        - export OPENMP_EXPECTED="True"
```
bsipocz: I still think we don't need any of this, setting them in the env part should be good enough, just as you do in the appveyor config

astropy_helpers/openmp_helpers.py
```diff
@@ -217,32 +237,34 @@ def add_openmp_flags_if_available(extension):
...
     return {return_bool}
 """[1:]
 
+
 def generate_openmp_enabled_py(packagename, srcdir='.'):
     """
-    Utility for creating `openmp_enabled.is_openmp_enabled()`
-    used to determine, post build, whether the package was built
-    with or without OpenMP support.
+    Generate `package.openmp_enabled.is_openmp_enabled`, which can then be used
```
bsipocz: double backticks here?

.travis.yml
```diff
@@ -70,8 +68,13 @@ before_install:
       # as exploratory tests for when the shipped version has automatic OpenMP support.
       # These tests will then fail and at such a time a new one should be added
       # to explicitly remove OpenMP support.
-    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export OPENMP_EXPECTED="False";
-      else export OPENMP_EXPECTED="True"; fi
+    - if [ -z $OPENMP_EXPECTED ]; then
```
bsipocz: Still, why do we need this at al? Variables defined in the env section are available to `os.environ` for later scripts.

astrofrog: @bsipocz - because we can't define it to be True for the default linux build and False for the default False build (builds defined in the ``env: matrix:`` section not the ``matrix: include:`` section).

astrofrog: Specifically for these builds:

```
 os:
     - osx
     - linux
 
 env:
   matrix:
     - PYTHON_VERSION=3.5
     - PYTHON_VERSION=3.6 SETUPTOOLS_VERSION=dev DEBUG=True
       CONDA_DEPENDENCIES='sphinx cython numpy pytest-cov'
```

bsipocz: ahh, OK, now I see.  I would have then just removed the second from here leaving only the egg build in the main matrix.

astrofrog: The first isn't an egg test so both would need to be taken out and four builds added to the include matrix, so I'm not sure it would simplify things that much. I'll go ahead and merge, but feel free to open a PR if you'd like to simplify things further.

bsipocz: ohh, indeed, should read more carefully (or stop multitask as pay more attention to the talk I'm sitting at...)

PR #396, opened 2018-06-19
Fix coverage reporting

astrofrog:
This must have changed in coverage or pytest-cov - now we need to include 'omit' in the reporting section too.


pllim:
Hmm does this need to be fixed over at template as well?

astrofrog:
For some reason I'm not seeing this issue with astropy, just here. I think it could be because for packages that use astropy-helpers things work fine, but astropy-helpers doesn't use astropy-helpers 😝

bsipocz:
hmm, we had those weird coverage dropping results for some generated c files lately, so I'm afraid it may be the same issue?

bsipocz:
e.g. for  https://github.com/astropy/astropy/pull/7058 here: https://coveralls.io/builds/17493870

pllim:
Tom, are you saying that we need a `astropy-helpers-helpers` sub-submodule? :wink:

bsipocz:
😱

astrofrog:
@bsipocz - ah I see, I don't have time to investigate but it might be worth trying a similar fix to this, i.e. having the same ``omit`` section in ``[run]`` and ``[report]``?

bsipocz:
Yeah, I don't have much time to investigate either, so suggest to run the experiment on CI. I'll open a PR momentarily for it, and we'll monitor whether those issues will come back or not.

bsipocz:
Actually those files are not in the omit list to start with, so this shouldn't be an issue there. 
Also, it's a bit difficult to judge actually which version is used, and also weird stuff is picked up e.g. numpy coverage? https://travis-ci.org/astropy/astropy/jobs/394470019#L1475

bsipocz:
(For reference the original issue was due to a change in coverage 4.5 to handle omit differently in run and report, and they just discovered that it was actually broken before anyway, so a fix in 4.5.1 was added: https://bitbucket.org/ned/coveragepy/src/d3f21a23416eae7ae6861a84753b71b06c0907b4/CHANGES.rst?at=default&fileviewer=file-view-default#CHANGES.rst-49)

PR #399, opened 2018-07-24
Extend exception handling of OpenMP support check

jamienoss:
``openmp_helpers.check_openmp_support()`` compiles and links the small C code to test
OpenMP support. If either of these stages fail, their exceptions are caught and
``is_openmp_supported = False`` is set. However, they may still pass but the
code fails to run. ``CalledProcessError`` also needs to be caught to deal with
this.


jamienoss:
Not really sure how to add a test for this.

I also continued with the silent approach as was the case with ``CompileError`` & ``LinkError``.

PR #410, opened 2018-10-16
Add functionality to disable OpenMP support

jamienoss:

### DIFF
--- a/astropy_helpers/openmp_helpers.py
+++ b/astropy_helpers/openmp_helpers.py
@@ -30,6 +30,13 @@ from .setup_helpers import get_compiler_option
 
 __all__ = ['add_openmp_flags_if_available']
 
+try:
+    # Check if this has already been instantiated, only set the default once.
+    _ASTROPY_DISABLE_SETUP_WITH_OPENMP_
+except NameError:
+    import builtins
+    # It hasn't, so do so.
+    builtins._ASTROPY_DISABLE_SETUP_WITH_OPENMP_ = False
 
 CCODE = """
 #include <omp.h>
@@ -233,6 +240,10 @@ def add_openmp_flags_if_available(extension):
     Returns `True` if the flags were added, `False` otherwise.
     """
 
+    if _ASTROPY_DISABLE_SETUP_WITH_OPENMP_:
+        log.info("OpenMP support has been explicitly disabled.")
+        return False
+
     openmp_flags = get_openmp_flags()
     using_openmp = check_openmp_support(openmp_flags=openmp_flags)
 
@@ -260,7 +271,7 @@ def is_openmp_enabled():
 """[1:]
 
 
-def generate_openmp_enabled_py(packagename, srcdir='.'):
+def generate_openmp_enabled_py(packagename, srcdir='.', disable_openmp=None):
     """
     Generate ``package.openmp_enabled.is_openmp_enabled``, which can then be used
     to determine, post build, whether the package was built with or without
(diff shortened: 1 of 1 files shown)


jamienoss:
This is in relation to https://github.com/astropy/astropy/pull/7293 and the request to use the environ variable, e.g. ``ASTROPY_OPENMP=1`` to conditional turn on and off OpenMP build support.

It would be used like it is here: https://github.com/jamienoss/astropy/pull/14, however, I don't like the gaping hole between the bool in the build generated func ``is_opnemp_eneabled()`` and the logic wrapping uses of ``add_openmp_flags_if_available()``. It's an issue for future dev, maintenance, and robustness. It is open to the scope of a module being compiled with OpenMP support yet ``is_opnemp_enabed()`` returns ``False`` and vice versa.

What I would like is some way to have a single entry point check that is then normalized to an internal representation, i.e. a global var or being able to call or read ``is_openmp_enabled()``, during the rest of the build and thus factorize the other checks into the other necessary openmp astropy-helpers funcs rather than multiple checks of ``ASTROPY_OPENMP``, which is already prone to obvious race conditions.

jamienoss:
@astrofrog FYI
Please let me know if there is a canonical way to use persistent variables during the build, I'm not distutils/build savvy to know.

astrofrog:
The main issue with making this too global is that we then have to make it opt-in or opt-out for ALL packages that use astropy-helpers. There may be some packages that already use astropy-helpers that are expecting it to just work and we'd break that by adding a global config. I think this is something that should probably be discussed in more detail with more developers. Since we're running out of time, for v3.1, I think that we should just comment out the call to ``add_openmp_flags_if_available`` in the core package and then figure things out for v3.2.

jamienoss:
Sorry, I probably didn't explain the situation well enough. This is really a current issue regarding the playoff between ``is_openmp_eneabled()`` which is generated at build time and then successive calls to ``add_openmp_flags_if_available()``. I'm saying that there exists already a slight issue, it's just not much of an issue until you explicitly want to disable OpenMP support.

We only want to read the environment variable once. This should be done at the beginning, i.e. passed to the call to ``generate_is_openmp_enabled()``. This then sets some flag that is persistent and visible to the rest of the build process. ``add_openmp_flags_if_available()`` would then read this and only add the flags conditionally on the state of this flag. The default for astropy-helpers would be for OpenMP to be on. There would then be no "opt-in by all", ``generate_is_openmp_enabled()`` wouldn't even need to be called, there would be zero impact on any other package that already adds the openmp flags with ``add_openmp_flags_if_available()``.

All that is missing is the correct pattern for this flag within the setup infrastructure. I don't know what that looks like, that's all.

The read of ``os.environ['ASTROPY_OPENMP']`` would only happen in astropy. It propagates this state via the ``disable_openmp`` param in ``generate_is_openmp_enabled()``. Other packages can do this however they want, with whatever environ variable they determine.

Does this make more sense now?

jamienoss:
The type of "global" that I think you are talking about already exists, it's the state returned from the generated function ``is_opnemp_enabled()``, but this is only robustly accessible post build & and installation. I need something that persists just during the build lifetime.

I'm not asking to make ``ASTROPY_OPENMP`` global.

astrofrog:
@jamienoss - I understand better what you mean now - I think the way we read the environment variable only once is that in astropy's setup.py we read it and then set a variable ``__USE_OPENMP__`` or something like this and add it to the builtins as we do for ``__ASTROPY_SETUP__``:

https://github.com/astropy/astropy-helpers/blob/master/astropy_helpers/__init__.py#L47

In the ``setup_package.py`` files we can then refer to ``__USE_OPENMP__``. Does this make sense? If so, then I think this PR is ready to merge, since the remaining changes would be in astropy core. Do you agree?

jamienoss:
Cool, thanks, I'll give that a whirl.

>since the remaining changes would be in astropy core. Do you agree?

No. I would need to add the check for this var within the other openmp helper code. _Then_ everything else would be in astropy.

jamienoss:
The only change then needed to astropy/setup.py would be:
```python 
...
# The current default is to NOT build with OpenMP support, i.e. ``disable_openmp = True``
disable_openmp = False if 'ASTROPY_SETUP_WITH_OPENMP' in os.environ and os.environ['ASTROPY_SETUP_WITH_OPENMP'] == '1' else True
generate_openmp_enabled_py(NAME, disable_openmp=disable_openmp)
...
```

I went with ``ASTROPY_SETUP_WITH_OPENMP``. I think the _"with"_ is esstential but just let me know if you'd prefer your suggestion of _"compile"_ rather than _"setup"_.

astropy_helpers/openmp_helpers.py
```diff
@@ -275,9 +286,16 @@ def generate_openmp_enabled_py(packagename, srcdir='.'):
     epoch = int(os.environ.get('SOURCE_DATE_EPOCH', time.time()))
     timestamp = datetime.datetime.utcfromtimestamp(epoch)
 
+    if disable_openmp != None:
```
astrofrog: Minor: this should be 'is not' rather than !=

astrofrog:
This looks good - just a minor comment above

PR #414, opened 2018-11-07
Switch to codecov for coverage reporting

astrofrog:
Ready for review!


astrofrog:
Results seem to agree with previously (and are slightly higher because we can now get coverage from AppVeyor)

PR #416, opened 2018-11-19
Added initial docs about what the package does

astrofrog:
I thought it was about time we documented what astropy-helpers actually does. This is pretty minimal, but I thought it'd be better than nothing.


mhvk:
Thanks for adding this! One possible addition that would be great -- but don't let it stand in the way of merging this! -- would be to tell also *where* these additions are done (i.e., link to the relevant files). In this respect, would it be crazy to run sphinx here too? Most functions seem very nicely documented. (On the other hand, if we're thinking of getting rid of this, then it would be a waste of time...)

saimn:
Maybe mention also `setup_package.py` files ?

cdeil:
@astrofrog - Thank you!

For me as affiliated package maintainer, the most interesting information would be how to update `astropy-helpers` and the related files we have inherited from the package-template, especially `setup.py` and `_astropy_init.py`.

So maybe add a mention of `_astropy_init.py`?

Also, if the bot is still running that will send PRs to do the updates, it should be mentioned as well?

There are other related docs pages, but they also don't give a clear path how to update:
http://docs.astropy.org/en/stable/development/astropy-package-template.html
http://docs.astropy.org/projects/package-template/en/latest/updating.html

Would one ever use `astropy-helpers` without the `_astropy_init.py` that comes from the package-template? Would you update the `astropy-helpers` always together with these few extra files that live outside the `astropy-helpers` folder or always in sync?

Basically: somewhere there should be a description of when / how to apply updates. I think that's the hard and time-consuming part for maintainers. Running cookie-cutter once is probably the easy part and already well-documented. So a better understanding / description how the few extra files from the package template relate to astropy-helpers (`_astropy_init.py`, `ah_bootstrap.py` and `setup.py` and `setup.cfg` mainly, I think) and how to update those would be most valuable for me.

astrofrog:
@cdeil - the ``_astropy_init.py`` file is mostly unrelated to astropy-helpers. I think the best place to put the instructions you mention would actually be the package-template docs: http://docs.astropy.org/projects/package-template/en/latest

mhvk:
I fear I have to second @cdeil here - The documentation for getting started with an affiliated package is really good but I've found keeping up to date tricky. Partially it is of course that it needs to be done only rarely, so all I remember is that I went by trial and error. And partially it is that I started before cookiecutter so have to do things manually. But partially also it is very difficult to know whether changes affects one's package.

One concrete recommendation is to provide links to changelogs - I know I found a summary of changes before (I think just on github), but the links above to not point to them.

Anyway, this is getting really a bit off-topic here!  And I fear I understand it too poorly to even help document it better -- which may be another good reason to try to minimize the number of moving parts.

README.rst
```diff
@@ -23,22 +24,117 @@ implementation are in the accepted
...
+
+* ``-l``: completely clean previous builds, including files generated by
+  the sphinx-automodapi package (which creates API pages for different
+  functions/classes).
+
+* ``-n``: disable the intersphinx option.
+
+* ``-o``: open the documentation in a browser if a build finishes successfully.
+
+In addition, build_docs will automatically download and temporarily install sphinx-astropy
```
cdeil: What does "temporarily install sphinx-astropy" mean?
Is it installed and removed again? In which environment / site-packages?
Does this always happen or only on first run?

README.rst
```diff
@@ -23,22 +24,117 @@ implementation are in the accepted
...
+directory, and run the tests using `pytest <http://pytest.org/>`_ on this
+installed version. Note that the bulk of this command is actually defined
+in ``astropy.tests.command.AstropyTest`` - here we simply define the custom
+setuptools command.
+
+**python setup.py sdist**
+
+We redefine ``sdist`` to use the version from distutils rather than from
+setuptools, as the setuptools version requires duplication of information
+in ``MANIFEST.in``.
```
cdeil: Is this customisation of `sdist` something that could maybe be removed completely?
I find `MANIFEST.in` confusing and don't understand what the good practices are concerning what include and exclude lines to put, i.e. whether to white or blacklist files, from the PYthon documentation.

So I can see why a modifcation was introduced here as an attempt to make it better.

But probably making it behave differently in Python packages that use astropy-helpers and "normal" Python packages that use setuptools isn't a good idea? It just creates more confusion overall?

bsipocz: see #419 for the discussion or removal of the custom `sdist`.

astrofrog:
Thanks all for your feedback!

@cdeil: I've added instructions on manually installing and updating astropy-helpers in a package. I've not added anything about _astropy_init.py since that's unrelated and package-template specific. I've also tried to explain better what temporary installation means.

@mhvk: documenting how to update other parts of packages using the package template is an issue for the package-template repo

@saimn: I've added a description of the ``setup_package.py`` files. This was actually present in the astropy docs hidden away in the developer docs so I stole some text from there:

http://docs.astropy.org/en/latest/development/building.html#customizing-setup-build-for-subpackages

I think it makes more sense for it to live here, so once this PR is merged, we can just add a link to this README from the astropy docs.

I think this is now a reasonably complete record of most things that can be done with astropy-helpers. Yes, it could live in sphinx docs, but let's see first what can be removed/simplified.

cdeil:
@astrofrog - Thank you very much. Having this description / reference is super useful. Maybe just merge this in now?

astrofrog:
@cdeil - I agree, I'll just wait on approvals from @eteq and @bsipocz as they are officially maintainers of the helpers.

README.rst
```diff
@@ -23,22 +30,304 @@ implementation are in the accepted
...
+------------------------
+
+If you would like to automatically get pull requests to update astropy-helpers,
+then see the instructions `here
+<https://github.com/astropy/astropy-procedures/tree/master/update-affiliated>`_.
+
+To instead update astropy-helpers manually, go inside the submodule and do::
+
+    cd astropy_helpers
+    git fetch origin
```
bsipocz: Most cases this is enough, but sometimes the tags are not coming through automatically with this line and I needed to add `--tags` to the fetch (e.g. it's definitely the case for the tag v3.1rc1). I don't understand the reason maybe something I messed up when releasing.

README.rst
```diff
@@ -23,22 +30,304 @@ implementation are in the accepted
...
+  (which are normally present in the stable releases) and give a nice error
+  if these are not found.
+
+Version helpers
+^^^^^^^^^^^^^^^^
+
+Another piece of functionality we provide in astropy-helpers is the ability
+to generate a ``packagename.version`` file that includes functions that
+automatically set the version string for developer versions, to e.g.
+``3.2.dev22213`` so that each developer version has a unique number. To use this,
```
bsipocz: This may be a bug or a feature, but people working on separate branches will share version numbers for different commits, so this uniqueness cannot be relied on on the package level.

README.rst
```diff
@@ -23,22 +30,304 @@ implementation are in the accepted
...
 `Affiliated package template <https://github.com/astropy/package-template>`_.
 
-.. image:: https://travis-ci.org/astropy/astropy-helpers.svg
-    :target: https://travis-ci.org/astropy/astropy-helpers
+Using astropy-helpers
+---------------------
+
+The easiest way to get set up with astropy-helpers in a new package is to use
+the `package-template <http://docs.astropy.org/projects/package-template>`_
+that we provide. However, we now go through the steps of adding astropy-helpers
```
eteq:
```suggestion
that we provide. This template is specifically designed for use with the helpers,
so using it avoids some of the tedium of setting up the heleprs.

However, we now go through the steps of adding astropy-helpers
```

README.rst
```diff
@@ -23,22 +30,304 @@ implementation are in the accepted
...
+Finally, add::
+
+    import ah_bootstrap
+
+at the top of your ``setup.py`` file. This will ensure that ``astropy_helpers``
+is now available to use in your ``setup.py`` file. Finally, add then commit your
+changes::
+
+    git add astropy_helpers ah_bootstrap.py setup.py
+    git commit ...
```
eteq:
```suggestion
    git commit -m "started using astropy-helpers"
```

README.rst
```diff
@@ -23,22 +30,304 @@ implementation are in the accepted
...
+
+Updating astropy-helpers
+------------------------
+
+If you would like to automatically get pull requests to update astropy-helpers,
+then see the instructions `here
```
eteq: This link doesn't seem to point to any actual instructions. What file actually has the instructions?  I suggest linking there instead, along with maybe some words like "If you would like the Astropy team to automatically generate pull requests..." - that is, make it clear that "automatically" really means "when there's a reason to do so" (assuming that's true, which I think it is, right?)

README.rst
eteq: I've seen the same thing and not fully understood why... but I think it might have to do with the the signing?  Or possibly *editing* of tags?

README.rst
```diff
@@ -23,22 +30,304 @@ implementation are in the accepted
...
+    git checkout v3.0.3
+
+Go back up to the root of the repository and update the ``ah_bootstap.py`` file
+too, then add your changes::
+
+    cp astropy_helpers/ah_bootstrap.py .
+    git add astropy_helpers ah_bootstrap.py
+    git commit ...
+
+What does astropy-helpers provide?
```
eteq: Maybe this section should come *first*?  I.e., before the "how to get it"?

README.rst
```diff
@@ -23,22 +30,304 @@ implementation are in the accepted
...
+           cmdclass=cmdclassd)
+
+The commands we provide or customize are:
+
+**python setup.py test**
+
+This command will automatically build the package, install it to a temporary
+directory, and run the tests using `pytest <http://pytest.org/>`_ on this
+installed version. Note that the bulk of this command is actually defined
+in ``astropy.tests.command.AstropyTest`` - here we simply define the custom
```
eteq: Maybe say *why* this is here?  (I.e. maybe:)
```suggestion
in ``astropy.tests.command.AstropyTest``, because that allows the test 
machinery to operate outside a setuptools command. This, here we 
simply define the custom
```

README.rst
```diff
@@ -23,22 +30,304 @@ implementation are in the accepted
...
+**python setup.py sdist**
+
+We redefine ``sdist`` to use the version from distutils rather than from
+setuptools, as the setuptools version requires duplication of information
+in ``MANIFEST.in``.
+
+**python setup.py build_docs**
+
+This command will automatically build the package, then run sphinx to build
+the documentation. Sphinx itself provides a custom setuptools command, which we
```
eteq:
```suggestion
the documentation. This makes development much easier because it ensures
sphinx extensions that use the package's  code to make documentation are 
actually using the in-development version of the code. Sphinx itself
provides a custom setuptools command, which we
```

README.rst
```diff
@@ -23,22 +30,304 @@ implementation are in the accepted
...
+  submodule that contains information about the compilers used.
+
+* Packages that need to build C extensions using the Numpy C API, we allow
+  those packages to define the include path as ``'numpy'`` as opposed to having
+  to import Numpy and call ``get_include``. The goal is to solve the issue that
+  if one has to import Numpy to define extensions, then Numpy has to be
+  installed/available before the package is installed, which means that one
+  needs to install Numpy in a separate installation step.
+
+* We detect broken compilers and replace them with other compilers on-the-fly
```
eteq: (reminder: remove this if #421 goes in, or remove it *in* #421 if this goes in first)

README.rst
```diff
@@ -23,22 +30,304 @@ implementation are in the accepted
...
+Go back up to the root of the repository and update the ``ah_bootstap.py`` file
+too, then add your changes::
+
+    cp astropy_helpers/ah_bootstrap.py .
+    git add astropy_helpers ah_bootstrap.py
+    git commit ...
+
+What does astropy-helpers provide?
+----------------------------------
+
```
eteq: Feel free to edit this, but I think a top level "what does it do big-picture" would help a bit.

```suggestion

Astropy-helpers' big-picture purpose is to provide customization to Python's
packaging infrastructure process in ways that the Astropy Project has found to
help simplifying the developint and releasing packages. This is primarily
build around ``setup.py`` commands, as outlined below, as well as code to help
manage version numbers and better control the build process of larger packages.

```

README.rst
bsipocz: My unfounded suspicion is that it happens when the tag is at the top of a branch? I'll try to remember to double check whether this is the case once the 3.1.x branch is updated.

bsipocz: just to confirm this is almost certainly an issue I've created by only pushing the tag back to github, but not the updated branch. Since the latter has been done I cannot reproduce it, the latest tag is available without doing `--tags`.

README.rst
astrofrog: Added

README.rst
astrofrog: I think this will go in first.

README.rst
astrofrog: I've re-ordered it

README.rst
astrofrog: Fixed

README.rst
astrofrog: I've tried to clarify this.

CHANGES.rst
```diff
@@ -4,7 +4,7 @@ astropy-helpers Changelog
 3.2 (unreleased)
 ----------------
 
-
+- Added extensive documentation about astropy-helpers to the README.rst file. [#416]
```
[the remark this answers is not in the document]
astrofrog: I've moved this to 3.1. I'm reluctant to move to 2.0.x since indeed it won't actually all be correct there.

astrofrog:
> I'm a bit worried about making sure this stays up-to-date... But eh, it's better to have it now and try to keep it coherent than not having it, given the complexity of this process.

One could say that about the entire astropy developer docs ;)

I've now implemented all the comments - since @eteq and @bsipocz approved this, I'll merge once the CI passes.

Issue #417, opened 2018-11-19
Consider removing _adjust_compiler


astrofrog:
The ``_adjust_compiler`` method provides a way of determining whether a 'broken' compiler is being used, and if so, to change it on-the-fly. In practice, this has only ever been used once, for a specific Mac version of gcc 4.2. In addition, while this compiler was broken for astropy.wcs, it doesn't mean it was for other packages.

In the spirit of trying to remove unnecessary functionality from astropy-helpers, I'd like to suggest that we remove this - first, this is an old issue and it's unlikely anyone is using this compiler now (even more so since most people install from conda/wheels anyway), and because if this happens in future, I feel like it makes more sense to document as a known issue than do magic with swapping compilers on-the-fly.

pllim:
I agree with removing it.

bsipocz:
I agree to remove it if we can, but disagree with "even more so since most people install from conda/wheels anyway", most of the helpers stuff is needed by us maintainers who work with and install from the source code.

hamogu:
+1 to removing it.

astrofrog:
Let's move the discussion to https://github.com/astropy/astropy-helpers/pull/421

astrofrog:
Done in #421

PR #421, opened 2018-11-23
Remove the adjust_compiler functionality

astrofrog:
This removes the ``adjust_compiler`` functionality, as initially discussed in https://github.com/astropy/astropy-helpers/issues/417. The motivation is that this has only ever been used for a single broken compiler (a specific version of gcc 4.2 on mac) and for a specific package (astropy.wcs) so it's no longer relevant and also was only ever really useful for the core package.

I think that if we ever have issues with specific compilers in future, we should just mention it in known issues.

⚠️ I think we should probably wait until the coordination meeting to decide on this ⚠️ 
(but feel feel to already approve or +1 if you agree)


CHANGES.rst
```diff
@@ -4,7 +4,10 @@ astropy-helpers Changelog
 3.2 (unreleased)
 ----------------
 
-
+- Remove functionality to adjust compilers if a broken compiler is detected.
+  This is not useful anymore as only a single compiler was previously patched
+  (now unlikely to be used) and this was only to fix a compilation issue in the
```
eteq: Is there an issue about the *original* problem?  If so it might be good to link to it *somewhere*, and this seems the logical place.

astrofrog: Looks like the original issue was https://github.com/astropy/astropy/issues/31 - I'll open a PR for a known issue.

astrofrog: Done - it's in the README now.

astrofrog:
@eteq - I added a known issue to the README.

PR #423, opened 2018-11-23
Updating numpydoc to v0.8.0

bsipocz:


pllim:
Huh... Any reason we still need `numpydoc` as opposed to [sphinx.ext.napoleon](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html)?

bsipocz:
Another conversation to have at the coordination meeting? (I vaguely remember it's an unresolved conversation).

bsipocz:
Oh, there you go, it started in 2015 if you follow the refs: https://github.com/astropy/sphinx-astropy/issues/5

eteq:
(see my comment back in the threads @bsipocz issued about how I tried Napolean and couldn't get it to work without spending a lot of time fixing things.  Still might be worth it, but it turns out it's not, at least, a simple drop-in)

PR #433, opened 2018-12-19
Fix get_compiler_option import

pllim:
I noticed that `get_compiler_option` is actually defined in `distutils_helpers.py`. You didn't get an error importing it from `setup_helpers.py` is because the latter imports it for its own use without cleaning its own namespace.


bsipocz:
Failure is unrelated, so I'm merging. 

Thank you @pllim!

PR #441, opened 2019-01-11
Include package name in error message for ah_bootstrap.py

astrofrog:
I decided to also move the parsing on setup.cfg to the top of the file which require reordering a bunch of stuff, hence the diff. With this, the error message looks like:

```
ERROR: Python (3, 6) or later is required by astropy-helpers for astropy
```

@drdavella @Cadair - is that clear enough?

### DIFF
--- a/ah_bootstrap.py
+++ b/ah_bootstrap.py
@@ -58,6 +58,9 @@ import pkg_resources
 from setuptools import Distribution
 from setuptools.package_index import PackageIndex
 
+# This is the minimum Python version required for astropy-helpers
+__minimum_python_version__ = (3, 5)
+
 # TODO: Maybe enable checking for a specific version of astropy_helpers?
 DIST_NAME = 'astropy-helpers'
 PACKAGE_NAME = 'astropy_helpers'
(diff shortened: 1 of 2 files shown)


bsipocz:
Somehow travis is not picking it up, we need to make sure it passes all builds there, too before merging.

astrofrog:
It's running here: https://travis-ci.org/astropy/astropy-helpers/pull_requests - maybe it will report once it's finished running (maybe the pending message got lost)

ah_bootstrap.py
```diff
@@ -45,18 +45,80 @@
...
+    ('auto_use', bool), ('path', str), ('download_if_needed', bool),
+    ('index_url', str), ('use_git', bool), ('offline', bool),
+    ('auto_upgrade', bool)
+]
+
+# Start off by parsing the setup.cfg file
+
+if os.path.exists('setup.cfg'):
+
+    SETUP_CFG = ConfigParser()
```
drdavella: Seems like this should be hoisted out of the `if`.

ah_bootstrap.py
```diff
@@ -45,18 +45,80 @@
...
+
+        log.error(
+            "Error reading setup.cfg: {0!r}\n{1} will not be "
+            "automatically bootstrapped and package installation may fail."
+            "\n{2}".format(e, PACKAGE_NAME, _err_help_msg))
+        SETUP_CFG = {}
+
+else:
+
+    SETUP_CFG = {}
```
drdavella: This shouldn't be necessary, nor the same statement above.

astrofrog:
@Cadair - technically speaking the requirement we are checking here is actually the one for astropy-helpers rather than for the package itself, hence why I wrote the error message this way.

Cadair:
Oh. Well that means sunpy would still need the check in the setup.py because we don't require the same version as helpers, so .... :man_facepalming:

bsipocz:
@Cadair - Oops. I suppose we can try to read in `minimum_python_version` from setup.cfg of the package (as well as from helpers)?

Cadair:
Yeah, that's what I was thinking, that we need two checks one for helpers and one for the package if we are having that in here.

astrofrog:
Ok I'll see what I can do!

astrofrog:
@Cadair @bsipocz - what about something like this now? It will first check if python_requires is present in ``setup.cfg`` and raise an appropriate error, e.g:

```
ERROR: Python >=3.6 is required by sunpy
```

then it will check if the version requirement for astropy-helpers itself is met. The reason I'm keeping the second test is because not all packages will define ``python_requires`` in ``setup.cfg``.

(Note that all this is of course irrelevant for recent versions of pip, which recognizes ``python_requires`` on the PyPI server before even downloading packages)

bsipocz:
@astrofrog - I'm very happy with this, thanks!

Issue #420, opened 2018-11-19
Consider if version helper is still needed


pllim:
https://github.com/spacetelescope/relic was written by @jhunkeler and it does automatic versioning through `git tag` functionality. I used it for some of my projects and find it rather convenient. Posting it here for your consideration.

bsipocz:
I think this is an excellent topic to bring up at the meeting.

hamogu:
I've used versioneer (https://pypi.org/project/versioneer/) in several package for this purpose and found it to work well, too.

astrofrog:
I looked into versioneer recently and it looked like it hadn't had any repo activity for a couple of years unfortunately :-/

saimn:
Another option is setuptools_scm, it looks pretty good and is developped by the pypa team:
https://github.com/pypa/setuptools_scm/

Cadair:
closed by APE 17 (it isn't)

PR #442, opened 2019-01-16
Avoid importing whole package when trying to get version information

astrofrog:
I'm currently working on trying to minimize/avoid importing packages during the setup process. This PR fixes two instances where we use ``__import__`` instead of ``import_file``, resulting in the package being imported just to get version information.


bsipocz:
Looking at the current diff, it looks like that this could be backported as is. Or this issue is not affecting the old approach?

astrofrog:
@bsipocz - yes it could potentially be backported, however the tests aren't passing and I think I'm going to have to rethink it a fair bit (I have some ideas).

astrofrog:
Ok so this got a bit more complicated than I had originally intended.

So to summarize, my main mission is to make it so that packages aren't imported during their own setup process because this is fragile and also very idiosyncratic to the Astropy project - most other projects (at least with C extensions) don't do this. The whole process of building with astropy-helpers is way over-complicated as a result of these idiosyncrasies.

One small piece of background: we provide a function called ``import_file`` in astropy-helpers which is just a wrapper around ``importlib.machinery.SourceFileLoader`` which allows specific files to be imported without importing the whole package.

During the setup process, we generate a ``version.py`` file as well as ``cython_version.py`` and a ``_compiler`` module. In the ``version.py`` file, we then import ``.cython_version`` and ``._compiler``. In Astropy core, there are a few places in the ``setup_package.py`` files where we try and import the ``version.py`` file. Currently in Astropy this automatically leads to the whole package getting imported because if you import ``astropy.version`` from a ``setup_package.py`` file it will import ``astropy``. Now we can start working around this by using ``import_file`` to import ``astropy/version.py`` - however, ``version.py`` includes imports to ``.cython_version`` and ``._compiler`` so that we still end up importing the whole package.

Now you might say 'well look here, the answer is simple, you can just use ``importlib.machinery.SourceFileLoader`` inside ``version.py`` to import ``cython_version.py`` and ``_compiler`` - however ``_compiler`` is not a source file, so we'd need to use ``ExtensionLoader`` instead, except that we don't know in advance what the filename of that extension will be (it will also be platform dependent) and we get into a mess of trying to glob the extension file, and so on and so forth.

An alternative is to simplify ``version.py`` so that it doesn't import ``.cython_version`` and ``._compiler``. This is the approach I've taken here - and I've renamed ``._compiler`` to ``.compiler_version``. The upside of this approach are that it's now possible to import e.g. ``astropy/version.py`` or ``astropy/cython_version.py`` without triggering an import of the whole package (e.g. astropy). The downside is that we remove e.g. ``astropy.version.cython_version`` and ``astropy.version.compiler`` from the API. The question is then just whether we think anyone is actually using these attributes in production anywhere. I'm going to guess no, but I might be missing something.

If we decide we really need to avoid breaking this API, then I can make the imports to ``cython_version`` and ``compiler_version`` inside ``version.py`` conditional on the value of ``_ASTROPY_SETUP_``, but I'd like to avoid that if possible since ideally we should be striving to remove the need for ``_ASTROPY_SETUP_`` in the long term.

TL;DR: I think this PR is ready but there will be some minor API breakage in the ``version`` module of packages that use astropy-helpers, specifically ``cython_version`` and ``compiler`` will no longer be defined.

pllim:
If we are going to break API anyway, why not go big and use https://github.com/pypa/setuptools_scm/ as suggested by @saimn in https://github.com/astropy/astropy-helpers/issues/420#issuecomment-455715777 ? 😉

CHANGES.rst
```diff
@@ -4,6 +4,8 @@ astropy-helpers Changelog
 3.2 (unreleased)
 ----------------
 
+- Avoid importing whole package when trying to get version information. [#442]
```
bsipocz: Please mention the minor api breakage of `cython_version` and `compiler` not being available any more

bsipocz:
@astrofrog - I'm OK with the API break. With a very quick github search I think you're right and no one really uses these from the version file, and anyway it's super minor so we can live with it. However now this is not backportable (which is OK, too).

astropy_helpers/version_helpers.py
```diff
@@ -358,11 +346,6 @@ def generate_version_py(packagename=None, version=None, release=None, debug=None
             f.write(_get_version_py_str(packagename, version, last_githash,
                                         release, debug, uses_git=uses_git))
 
-        invalidate_caches()
-
-        if version_module:
-            imp.reload(version_module)
```
eteq: I don't understand why it's safe to delete this.  Isn't this necessary to ensure that the version actually gets imported if there was a previous file already there?

astrofrog: Basically this isn't needed anymore because we import the version module with ``import_file`` rather than using ``__import__`` - so we aren't *actually* importing ``astropy.version`` (and therefore ``astropy``) anymore and don't need to reload it.

embray:
Has there been any further look at using setuptools_scm?  I just recently integrated it into a new project, and it's come a long way since I last used it.  It still took a little more work than I'd hoped to get it to do exactly what I wanted, but in the end I was pretty pleased with the results.

astrofrog:
@embray - we are using setuptools_scm now for astropy core and other astropy packages (see APE 17). This repository is now in bugfix-only mode so I don't think there's any point in refactoring the version-handling stuff here.

embray:
Oh, Good!  I missed APE 17 ^^

PR #443, opened 2019-01-17
Create setup() wrapper that does all the standard astropy-helpers actions

astrofrog:
With these changes, most packages that use astropy-helpers will be able to reduce their ``setup.py`` file to:

```python
import ah_bootstrap
from astropy_helpers.setup_helpers import setup
setup()
```

I'm not suggesting removing the individual functions, but just making this standard case easier. I'm going to write up some documentation about this too, but reviews of the implementation are welcome.

I'll add the docs here once #444 is merged

### DIFF
--- a/astropy_helpers/setup_helpers.py
+++ b/astropy_helpers/setup_helpers.py
@@ -14,6 +14,7 @@ import sys
 import traceback
 import warnings
 from configparser import ConfigParser
+import builtins
 
 from distutils import log
 from distutils.errors import DistutilsOptionError, DistutilsModuleError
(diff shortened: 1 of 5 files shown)


astrofrog:
The CircleCI failure can be ignored, it is because the configuration is only added in #444

astropy_helpers/setup_helpers.py
```diff
@@ -67,6 +70,39 @@
...
+    """
+    A wrapper around setuptools' setup() function that automatically sets up
+    custom commands, generates a version file, and customizes the setup process
+    via the ``setup_package.py`` files.
+    """
+
+    # Store the package name in a built-in variable so it's easy
+    # to get from other parts of the setup infrastructure
+    conf = read_configuration('setup.cfg')
+    builtins._ASTROPY_PACKAGE_NAME_ = conf['metadata']['name']
```
Cadair: Still not clear on what this does...

bsipocz:
Looks good to me, but as we discussed on slack I would like to take this for a ride with multiple coordinated packages. It can be done after the merge though (a bit easier to test), and if needed any adjustments can be done in a follow-up PR.

PR #444, opened 2019-01-17
Move documentation to Sphinx

astrofrog:
This also includes a docs build on CircleCI with a link provided by Giles


astrofrog:
Just confirming this also works properly on RTD: https://astropy-helpers.readthedocs.io/en/latest/

docs/conf.py
```diff
@@ -0,0 +1,52 @@
+# -*- coding: utf-8 -*-
+
+project = 'astropy-helpers'
+copyright = '2019, The Astropy Developers'
```
bsipocz: what's the policy on the year here? I mean it keeps coming up every year, and I keep forgetting. Should it be the full time period, or the latest or the first year?

bsipocz:
I liked the readme way of documenting infrastructure like the helpers, and also not a big fan of circleCI, but agreed that the readme was already too long.

astrofrog:
The reason I did this is because I needed to add more docs and this was getting very long. I'm going to add some of the high-level description back to the README though re: @eteq's comment.

docs/conf.py
astrofrog: For now I'll fix it to make it match the license, but we can update this if we figure out what the right one is.

bsipocz: Oh, sure. This was more like self mumbling that we need to figure out and act on separately across the whole ecosystem. (preferably with batchpr? 😉  )

astrofrog: Ok cool :)

astrofrog:
I've backported (and updated) this to v3.1.1

PR #445, opened 2019-01-17
Moved most of the setup.py configuration to setup.cfg

astrofrog:
Also minimize how much we use astropy-helpers to install astropy-helpers (still keeping the version adjustment but nothing else)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
-# NOTE: most of the configuration is defined in setup.cfg
+# NOTE: most of the configuration, including the version number,
+# is defined in setup.cfg
 
 import sys
 from distutils.version import LooseVersion
--- a/setup.py
+++ b/setup.py
@@ -1,54 +1,19 @@
 #!/usr/bin/env python
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
-import ah_bootstrap
-import pkg_resources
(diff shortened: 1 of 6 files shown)


astrofrog:
The CircleCI failure can be ignored, it is because the configuration is only added in #444

setup.py
```diff
@@ -1,53 +1,19 @@
 #!/usr/bin/env python
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
-import ah_bootstrap
-import pkg_resources
-from setuptools import setup
-from astropy_helpers.setup_helpers import (register_commands, get_package_info,
-                                           add_exclude_packages)
-from astropy_helpers.version_helpers import generate_version_py
+# NOTE: most of the configuration is defined in setup.cfg
```
eteq:
```suggestion
# NOTE: most of the configuration, including the version number, 
# is defined in setup.cfg
```

because I think the most likely case here is someone looking here and wondering "how do I update the version?"

bsipocz:
All looks good, enormous thank you @astrofrog for doing all these cleanups for the helpers.

Also a note to myself, I need to learn/get used to the new `[options.package_data]` way instead of `setup_package.py` file, it's so much neater.

astrofrog:
> Also a note to myself, I need to learn/get used to the new [options.package_data] way instead of setup_package.py file, it's so much neater.

And I think we should actually recommend that for the package template, it's overkill to use ``setup_package.py`` for anything other than the core package (and maybe e.g. Sunpy)

PR #446, opened 2019-01-20
Simplify custom build_ext

astrofrog:
As the code itself warned, this was previously spaghetti code 🍝 

It looks like this can be simplified - I guess things were different before the distribute/setuptools merger perhaps.


astrofrog:
I introduced a subtle test failure in my last commit, will investigate shortly.

embray:
I haven't really looked over it carefully either, but a lot of the crud in that class was previously in service to older versions of setuptools that did a lot of nonsense with Cython that I think has now been removed.  

I would just make sure this works well in the case where Cython is not installed, but the Cythonified C sources are still available and buildable.

astrofrog:
I just tried downloading the regions tar file release and installing it using this version of astropy-helpers, with no Cython installed, and it worked fine.

Now in the regions package, the extensions aren't actually explicitly defined:

https://github.com/astropy/regions/tree/master/regions/_geometry

There are pyx files but they somehow get auto-identified as extensions by setuptools. So if I remove the .pyx files and then keep only the generated C files, the extensions don't get picked up and the package is not build correctly. However, note that this is also the case before this PR - this really just refactors some code but doesn't actually change any initial assumptions.

I thought we used to ship some packages without the Cython code, just the generated C code, so I'm a little confused as to how that ever worked... In any case, as indicated above, this is not something that's changed by this PR.

astrofrog:
Ah right sorry that behavior comes from astropy-helpers' own ``get_cython_extensions``. So it looks like we have designed it specifically so that auto-discovery works only if the pyx files are present, so that all makes sense and is unchanged here.

astropy_helpers/commands/build_ext.py
```diff
@@ -1,329 +1,180 @@
...
         return False
 
 
-_compiler_versions = {}
-
-
-# TODO: I think this can be reworked without having to create the class
-# programmatically.
-def generate_build_ext_command(packagename, release):
+class AstropyBuildExt(SetuptoolsBuildExt):
```
eteq: It might confuse people that this is the "astropy builder" given that it's really not astropy itself...

```suggestion
class HelpersBuildExt(SetuptoolsBuildExt):
```
(of course accepting this would also require changing the names below as well...)

astropy_helpers/commands/build_ext.py
```diff
@@ -1,329 +1,180 @@
...
+            self.previous_cython_version = 'unknown'
+
+        self.uses_cython = should_build_with_cython(self.previous_cython_version, self.is_release)
+
+        # Add a copy of the _compiler.so module as well, but only if there
+        # are in fact C modules to compile (otherwise there's no reason to
+        # include a record of the compiler used). Note that self.extensions
+        # may not be set yet, but self.distribution.ext_modules is where any
+        # extension modules passed to setup() can be found
+        extensions = self.distribution.ext_modules
```
eteq: Maybe this should go *above* the comment block, since the comment actually applies for the below part more directly?

astropy_helpers/commands/build_ext.py
```diff
@@ -1,329 +1,180 @@
...
-            modules are present (if building without Cython installed).
-            """
-
-            # Determine the compiler we'll be using
-            if self.compiler is None:
-                compiler = get_default_compiler()
-            else:
-                compiler = self.compiler
+    uses_cython = False
+    force_rebuild = False
```
eteq: I imagine at this point the answer is "it's a breaking change"... but isn't there alse a ``--force`` option?  Is that different from ``--force-rebuild``?  If not, perhaps a deprecation is called for?

astropy_helpers/commands/build_ext.py
```diff
@@ -1,329 +1,180 @@
...
+            if 'numpy' in extension.include_dirs:
+                if np_include is None:
+                    np_include = get_numpy_include_path()
+                idx = extension.include_dirs.index('numpy')
+                extension.include_dirs.insert(idx, np_include)
+                extension.include_dirs.remove('numpy')
+
+            self._check_cython_sources(extension)
+
+        # Note that setuptools automatically uses Cython to doscover and
```
eteq:
```suggestion
        # Note that setuptools automatically uses Cython to discover and
```

astropy_helpers/commands/build_ext.py
```diff
@@ -1,329 +1,180 @@
...
-
-    return build_ext
+                    msg = (
+                        'Could not find C/C++ file {0}.(c/cpp) for Cython '
+                        'file {1} when building extension {2}. Cython '
+                        'must be installed to build from a git '
+                        'checkout.'.format(base, pyxfn, extension.name))
+                    raise IOError(errno.ENOENT, msg, cfn)
+
+            # Current versions of Cython use deprecated Numpy API features
```
eteq: I was going to suggest deleting this, but then tried it and the warnings in question are still there...

```suggestion
            # Cython (at least as of 0.29.2) uses deprecated Numpy API features
```

bsipocz:
> One thing that would be good if possible, though: can we determine the earliest version of setuptools this now works for and set that as the minimum setuptools version?

We already require setuptools 30.3, and testing it on CI, so we've been covered I think.

astropy_helpers/commands/build_ext.py
```diff
@@ -1,329 +1,180 @@
...
 
-from ..utils import get_numpy_include_path, classproperty
-from ..version_helpers import get_pkg_version_module
+from ..distutils_helpers import get_main_package_directory
+from ..utils import get_numpy_include_path, import_file
 
 
-def should_build_with_cython(package, release=None):
-    """Returns the previously used Cython version (or 'unknown' if not
+def should_build_with_cython(previous_cython_version, is_release):
```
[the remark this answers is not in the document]
astrofrog: This function got simplified a little in the refactoring, and all 33 instances of this function on GitHub are just from astropy-helpers itself (i.e. it's never called externally). I think it's safe to change the API here.

astropy_helpers/commands/build_ext.py
astrofrog: I was mostly going with the naming convention used for other builders, e.g. ``AstropyBuildDocs`` but I see your point. I might change it to ``AstropyHelpersBuildExt`` instead of ``HelpersBuildExt`` though?

astropy_helpers/commands/build_ext.py
astrofrog: In fact ``force_rebuild`` is mostly an internal flag in this class that isn't actually needed as a command-line option. I'll make it private and I don't think deprecation is needed since ``--force`` has always been the correct flag to use.

astropy_helpers/commands/build_ext.py
astrofrog: The ``extensions = ...`` is actually part of the block below, but I've removed the empty line below because I agree it looked confusing before. Now it's all one block of code it should be clearer

astrofrog:
@bsipocz is right - the version at which setuptools started to deal with Cython correctly is quite a bit older than 30.3, which is our current minimum. It might even be before setuptools 1.x from what I can tell.

PR #447, opened 2019-01-20
Remove code from astropy_helpers.utils that isn't used

astrofrog:
These utilities aren't used anymore - and if we do need to deprecate things, the code base for astropy-helpers is now small enough that raising the deprecation warning manually would be more efficient.


astrofrog:
Huh I wonder why the astropy bot didn't do anything here

astrofrog:
About backporting to LTS: true but I think we shouldn't since this is formally an API change/removal.

bsipocz:
> we shouldn't since this is formally an API change/removal.

we didn't really apply this rule for the helpers.

bsipocz:
Thanks @astrofrog!

pllim:
From the comment, it sounded like the original plan was to move these from `astropy` core to `astropy-helpers`? I can see why as these are generic non-astro functions. But I also have no seen any user request to do so. No (belated) objection here.

astrofrog:
Note that astropy-helpers is only available at setup time, so we couldn't ever actually remove these from astropy core.

pllim:
Maybe back in 2015, this wasn't a submodule yet? Ah, well. :woman_shrugging:

PR #449, opened 2019-01-20
Remove mention of get_entry_points which isn't actually implemented

astrofrog:
Just removing docs about a feature that doesn't exist.


bsipocz:
Good catch! Thanks @astrofrog!

PR #450, opened 2019-01-28
Avoid importing astropy when getting the test command to test astropy

astrofrog:
In my cleanup series, this is the last place where astropy-helpers causes astropy to be imported during setup (there are some changes on the astropy side that are needed to avoid this too). I think it makes more sense to use ``import_file`` to just import the command file here.

### DIFF
--- a/astropy_helpers/commands/test.py
+++ b/astropy_helpers/commands/test.py
@@ -2,14 +2,10 @@
 Different implementations of the ``./setup.py test`` command depending on
 what's locally available.
 
-If Astropy v1.1.0.dev or later is available it should be possible to import
-AstropyTest from ``astropy.tests.command``.  If ``astropy`` can be imported
-but not ``astropy.tests.command`` (i.e. an older version of Astropy), we can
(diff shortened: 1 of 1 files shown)


astropy_helpers/commands/test.py
```diff
@@ -2,25 +2,33 @@
...
 # fail
+
 try:
-    import astropy  # noqa
-    from astropy.tests.command import AstropyTest
+
+    # If we are testing astropy itself, we need to use import_file to avoid
+    # actually importing astropy (just the file we need).
+    command_file = os.path.join('astropy', 'tests', 'command.py')
+    if os.path.exists(command_file):
```
bsipocz: would this work, too when running `astropy.test()` or it doesn't matter at that point anyway as the helpers is not installed?

astrofrog: ``astropy.test()`` doesn't use the test command (just the test runner directly)

bsipocz:
Thanks @astrofrog!

PR #451, opened 2019-02-03
Write error messages on stderr

dpshelio:
Some messages were using `file` on `print`, whereas other where just simple
`print`s. Now all are using `sys.stderr.write`.


bsipocz:
Thanks @dpshelio!

PR #453, opened 2019-02-11
Preserve package data defined in [options.package_data]

astrofrog:
Currently, astropy-helpers's ``get_package_info`` function ignores package data defined in ``setup.cfg``. This fixes things so that we start off from what's defined in ``setup.cfg`` then iterate through the ``setup_package.py`` files. Not really worth backporting since including data in ``[options.package_data]`` hasn't been standard practice for us so far (though I hope to change that).


bsipocz:
Thank you @astrofrog!

PR #458, opened 2019-04-03
adding sphinx-dev testing to the cron jobs

bsipocz:
The sphinx-dev issue should have been picked up sooner that the cron run that job.


bsipocz:
The sphinx-dev test is failing, but I'm merging as this PR doesn't make the situation worse (and it's an issue on master already).

Issue #462, opened 2019-04-23
astropy-helpers should respect python_requires


bsipocz:
Matplotlib stopped supporting python3.5 in their new release (it's not a release just yet, but 3.1rc1). This causes failure on our CI here for the python3.5 build.

However this our own fault as the machinery here doesn't respect the `python_requires` in mpl's setup.py.

@tacaswell kindly digged up that the offending lines are likely  https://github.com/astropy/astropy-helpers/blob/b17db72b5b8b039ffb42a767cc8311f6f1c77b49/astropy_helpers/commands/build_sphinx.py#L58-L60 

```
from setuptools import Distribution
dist = Distribution()
eggs = dist.fetch_build_eggs('sphinx-astropy')
```

I'll work this around for our CI to get the RC out sooner rather than later, but it would be great if we fix this before the actual release (and it may need to be backported, to LTS, too).

astrofrog:
I think this is because setuptools doesn’t recognise python_requires. Not sure if we can easily use pip instead though.

bsipocz:
Or give up on seamlessly install sphinx-astropy and ask packages to explicitly list it (as they need to do it with pytest, sphinx, etc.)

PR #464, opened 2019-04-23
Make sure we directly install mpl for py3.5

bsipocz:
This is to workaround #462, as I couldn't find the way to make setuptools understand `python_require` (nor pre-release, but that's an issue already been reported upstream: https://github.com/pypa/setuptools/issues/855).

There is another instance of `fetch_build_eggs()` in `ah_bootstrap.py`, but I leave that one alone for now as packages listing their dependencies should know what

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,10 +45,10 @@ distribute-*.tar.gz
 .coverage
 cover
 htmlcov
+.pytest_cache
 
 # Mac OSX
 .DS_Store
 
 # PyCharm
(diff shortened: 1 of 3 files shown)


bsipocz:
The are separate issues with numpydoc requiring sphinx 1.6+, but I'll follow-up on that in a separate PR.

bsipocz:
This is an uncontroversial workaround, so I'm merging it now.

PR #466, opened 2019-04-23
OPENMP_EXPECTED defaults to False silently

### LINKED ISSUE #460: Default OPENMP_EXPECTED to False
Locally this is very annoying, I strongly suggest to default to False if the variable doesn't exist rather than fail.

```
======================================================== ERRORS ========================================================
____________________________ ERROR collecting astropy_helpers/tests/test_openmp_helpers.py _____________________________
astropy_helpers/tests/test_openmp_helpers.py:20: in <module>
    OPENMP_EXPECTED = 'True' == os.environ['OPENMP_EXPECTED']
../../../.pyenv/versions/3.5.6/lib/python3.5/os.py:725: in __getitem__
    raise KeyError(key) from None
E   KeyError: 'OPENMP_EXPECTED'

During handling of the above exception, another exception occurred:
astropy_helpers/tests/test_openmp_helpers.py:22: in <module>
    raise Exception("The OPENMP_EXPECTED environment variable should be set to "
E   Exception: The OPENMP_EXPECTED environment variable should be set to True or False and should indicate whether OpenMP should work on your platform.
```

pllim:
Fix #460


bsipocz:
Thanks @pllim! There are failures expected that are unrelated, so I'll merge this even if CI is in red with those.

PR #467, opened 2019-04-23
Clean up openmp test file

### LINKED ISSUE #461: openmp_enabled.py should be a temp file rather than in the repo
Running tests dumps a `openmp_enabled.py` file in the main repo directory. It should rather go into a temporary file that gets deleted after the tests are run.

Content:
```
# Autogenerated by 's setup.py on 2019-04-23 00:18:30

def is_openmp_enabled():
    """
    Determine whether this package was built with OpenMP support.
    """
    return False`
```

pllim:
Fix #461

### DIFF
--- a/astropy_helpers/tests/test_openmp_helpers.py
+++ b/astropy_helpers/tests/test_openmp_helpers.py
@@ -50,25 +50,35 @@ def test_add_openmp_flags_if_available():
         assert not using_openmp
 
 
-def test_generate_openmp_enabled_py():
+class TestOpenMPEnabled:
+    @classmethod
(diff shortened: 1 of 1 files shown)


pllim:
Alternate way if to retain the test as a function (instead of turning it into a class) but put the whole thing under `try-finally` block, but I prefer this way personally.

bsipocz:
Failure should be related, I see it locally, too.

pllim:
Weird, it passed when I ran with `pytest testfilename`...

pllim:
Oh... It is failing in another test file... Hmm.

PR #468, opened 2019-04-23
Ignore .coverage.subprocess

pllim:
I keep seeing this as untracked file after I run the tests locally.


bsipocz:
Thanks @pllim!

PR #471, opened 2019-04-23
Install Pillow with conda to avoid compilation from source

saimn:
Pillow comes as a dependency of sphinx-astropy, which itself is
installed under the hood with setuptools (to test that this installation
mode works). Which means that Pillow is compiled from source, which is
wasting some resources for nothing.


bsipocz:
@saimn - The builds in this PR are not quicker, and in the meantime Pillow and the rest of the dependencies are now only installed as eggs in one of the builds. I would vote to keep that around, but am also strongly in favour or removing the whole sphinx-astropy install as egg workaround and say explicitly that anyone relying on the helpers to build their docs should have it as a dependency.

Issue #472, opened 2019-04-24
Make sphinx-astropy an explicit dependency for docs build


bsipocz:
Packages relying on astropy-helpers to provide docs building infrastructure should make sphinx-astropy an explicit dependency. Being cleaver and building it and the full dependency three as eggs proved to be problematic when those dependencies start to introduce some version restrictions (e.g. mpl for python, and numpydoc for sphinx) See e.g., #462, #463, #465, etc.

And in addition building the eggs for all those dependencies is the least efficient way to install them (#471)

Now it is a good time to make this change, before the actual v3.2 release.

cc @astrofrog

astrofrog:
I personally agree, I think it's reasonable to just crash and tell users that sphinx-astropy is needed.

pllim:
Where exactly do you want that message? There is already this here

https://github.com/astropy/astropy-helpers/blob/222e6b5cdb62cdecf8221f3f4c70fa82beda83fe/astropy_helpers/sphinx/conf.py#L5

astrofrog:
I think we should also raise an error here:

https://github.com/astropy/astropy-helpers/blob/master/astropy_helpers/commands/build_sphinx.py#L192

just to be on the safe side (instead of calling ``ensure_...``.

bsipocz:
I'll open a PR for this at the end of the today.

saimn:
And we also have now an explicit dependency with sphinx-asdf, so installing sphinx-astropy under the hood is not enough to build the docs anyway.

astrofrog:
Arguably we could make sphinx-asdf a dependency of sphinx-astropy, but let's fix its major performance issues first!

bsipocz:
This was really done for packages other than astropy who use the template and the helpers but not necessarily know what goes on under the hood. astropy always had the explicit dependency on `sphinx-astropy` listed in setup.py/setup.cfg.

bsipocz:
Addressed by #474 (and now I realized "address" is not a magic word, so that PR didn't close this issue).

PR #474, opened 2019-05-02
Remove auto installing sphinx astropy

### LINKED ISSUE #462: astropy-helpers should respect python_requires

### LINKED ISSUE #463: Prereleases should not be picked up automatically
In our CI we pick up docs dependencies and egg install them. 

However even then we should not pick up pre-releases.

Related build (that fails because of #462): https://travis-ci.org/astropy/astropy-helpers/jobs/523215037#L1194

bsipocz:
This is to address #472 and all underlying issues listed in there.

Fix #462 fix #463

### DIFF
--- a/astropy_helpers/commands/build_sphinx.py
+++ b/astropy_helpers/commands/build_sphinx.py
@@ -6,7 +6,6 @@ import re
 import shutil
 import subprocess
 import sys
-import glob
 import warnings
 from distutils.version import LooseVersion
 
@@ -39,7 +38,7 @@ for builder in {builders!r}:
 
 def ensure_sphinx_astropy_installed():
     """
-    Make sure that sphinx-astropy is available, installing it temporarily if not.
+    Make sure that sphinx-astropy is available.
 
     This returns the available version of sphinx-astropy as well as any
     paths that should be added to sys.path for sphinx-astropy to be available.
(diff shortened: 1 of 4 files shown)


astropy_helpers/commands/build_sphinx.py
```diff
@@ -55,36 +54,9 @@ def ensure_sphinx_astropy_installed():
...
-                if egg.project_name == 'sphinx-astropy':
-                    sphinx_astropy_version = egg.parsed_version.public
-                    break
-
-        eggs_path = os.path.abspath('.eggs')
-        for egg in glob.glob(os.path.join(eggs_path, '*.egg')):
-            sys_path_inserts.append(egg)
+        raise ImportError("sphinx-astropy is required to build documentation "
+                          "using the astropy machinery. Please install the "
+                          "package and try again.")
```
astrofrog: What about just 'sphinx-astropy needs to be installed to build the documentation' to keep it short?

bsipocz: done

bsipocz:
I go ahead and merge this to tag an rc2

PR #475, opened 2019-05-03
Removing sphinx extension that works around old bug 

bsipocz:
Issues with the extension causing docs build failure on LTS astropy. I don't think it's worth spending time on fixing it, rather do the same as in https://github.com/astropy/sphinx-astropy/commit/16065ccac6cb758ae29b27e02de8f36fc45cc603


bsipocz:
CircleCI is not set up for this branch, so ignore the status.

PR #480, opened 2019-05-30
Remove PY2 compat for C compiler

pllim:
We don't need to check whether it is PY2 or PY3 anymore, right?

### DIFF
--- a/astropy_helpers/commands/src/compiler.c
+++ b/astropy_helpers/commands/src/compiler.c
@@ -7,15 +7,6 @@
  * the compilers we care about.
  ***************************************************************************/
 
-#ifndef PY3K
-#if PY_MAJOR_VERSION >= 3
-#define PY3K 1
-#else
-#define PY3K 0
-#endif
-#endif
-
-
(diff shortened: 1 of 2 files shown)


bsipocz:
Thanks @pllim, indeed.

bsipocz:
A wonder why the codecov bot behaves here, but double comment on the astropy repo.

Anyway, this is all green now, so I'm merging. Thanks again @pllim!

pllim:
Maybe because I changed the C stuff, it doesn't know how to calculate? 🤷‍♀

bsipocz:
I don't mean the content of the comment, just that fact that finally it doesn;'t comment twice. Not sure why it does it for astropy core, and I also heard it really badly behaves for other repos, e.g. e.g the bot is muted for numpy due to the repeated commenting.

pllim:
There was a similar behavior from `astropy-bot` before. I think @astrofrog said it was race condition when posting comment from multiple events that happen too closely in time. That was why we moved away from bot posting a comment, and now it posts status.

Issue #481, opened 2019-06-05
remove deprecation for build_sphinx  command


bsipocz:
`build_sphinx` is still the standard way for sphinx setuptools integration. I strongly feel that we should remove the deprecation warning and keep it along with the convenience `build_docs`.

pllim:
Wasn't there a big discussion when we decided to go with `build_docs`? @eteq always says there should only be one way to do things. 🤔

bsipocz:
Yes, but then `build_sphinx` is the canonical way to do it.

eteq:
A bit of googling implies to me that `build_docs` is still only used pretty much only in the astropy community, whereas `build_sphinx` is the standard elsewhere.  So on that basis I think I'm 👍 to this.  To truly get to the "only one way to do it" at this point I think we would need a bigger break like astropy/astropy-APEs#52 ...

(Out-of-band @pllim revealed that it was me that argued for `build_docs` in the first place?  But @Cadair agreed so it must have been a good idea at the time 😜 )

pllim:
Just gonna xref some history that I dug up related to this during the coworking hour today:

* astropy/astropy#5157
* astropy/astropy#3924
* astropy/astropy-helpers#161

Personally, I am okay with @bsipocz 's idea to remove the warning.

Cadair:
`import this`...

bsipocz:
The removal of deprecation warning has been merged in, feel free to reopen this issue if you prefer a bigger cleanup.

PR #482, opened 2019-06-06
Remove deprecation warning for build_sphinx

bsipocz:
To address #481


astropy_helpers/tests/test_setup_helpers.py
```diff
@@ -309,8 +309,6 @@ class B(A):
             run_setup('setup.py', ['build_docs', '-l'])
         elif mode == 'deprecated':
```
pllim: Nitpick: Do we want to rename this mode to something not "deprecated" since it is no longer deprecated?

bsipocz: good point, I haven't noticed it.

bsipocz:
Remilestoning to 3.2.1 as I don't feel like releasing 2.0.11 now to include only this one change.

Issue #487, opened 2019-06-19
AttributeError: type object '_DummyCommand' has no attribute 'error_msg'


cdeil:
I get this error:
```
(regions) hfm-1804a:regions deil$ pytest
====================================================================================== test session starts ======================================================================================
platform darwin -- Python 3.7.3, pytest-4.3.1, py-1.8.0, pluggy-0.9.0
rootdir: /Users/deil/temp/regions, inifile: setup.cfg
plugins: remotedata-0.3.1, openfiles-0.3.2, doctestplus-0.3.0, cov-2.7.1, arraydiff-0.3
collected 1031 items / 1 errors / 1030 selected                                                                                                                                                 

============================================================================================ ERRORS =============================================================================================
______________________________________________________________ ERROR collecting astropy_helpers/astropy_helpers/commands/_dummy.py ______________________________________________________________
astropy_helpers/astropy_helpers/commands/_dummy.py:69: in __getattribute__
    raise DistutilsArgError(cls.error_msg)
astropy_helpers/astropy_helpers/commands/_dummy.py:67: in __getattribute__
    return super(_DummyCommandMeta, cls).__getattribute__(attr)
E   AttributeError: type object '_DummyCommand' has no attribute 'error_msg'
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
==================================================================================== 1 error in 5.31 seconds ====================================================================================
```
Log: https://gist.github.com/cdeil/aeae63805d091db7cba64e98e1360697

You should be able to reproduce with:
```
git clone https://github.com/astropy/regions.git
cd regions/
pip install -e .
pytest
```
in a Python 3.7 env where the relevant dependencies (including pytest) are available.

astrofrog:
Can you post the output of ``pip freeze`` too?

pllim:
Use this command instead:
```
python setup.py test
```

astrofrog:
Or alternatively:

     pytest regions

cdeil:
Sure, there's ways to avoid the error. I reported it because I thought it's a regression / but you'd want to fix. If you don't think it's worth it, feel free to just close this issue.

[...]

PR #488, opened 2019-06-19
Fix error that occurred when using bare pytest in packages that use astropy-helpers

astrofrog:
Note that I think this should be fixed *anyway* because otherwise it will cause issues for any kind of package that tries to do some automatic introspection on ``_DummyCommand``.

Fixes https://github.com/astropy/astropy-helpers/issues/487

@cdeil - ideally we should probably adjust the package template so that ``setup.cfg`` includes options for pytest so that it ignores ``astropy_helpers`` by default. If you know what flag that should be, could you open a PR to the package-template?

### DIFF
--- a/astropy_helpers/commands/_dummy.py
+++ b/astropy_helpers/commands/_dummy.py
@@ -61,7 +61,7 @@ class _DummyCommandMeta(type):
                 "attribute.")
 
     def __getattribute__(cls, attr):
-        if attr in ('description', 'error_msg'):
+        if attr in ('description', 'error_msg') or attr.startswith('_'):
             # Allow cls.description to work so that `./setup.py
             # --help-commands` still works
             return super(_DummyCommandMeta, cls).__getattribute__(attr)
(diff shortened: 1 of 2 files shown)


astropy_helpers/commands/_dummy.py
```diff
@@ -61,7 +61,7 @@ def __init__(cls, name, bases, members):
                 "attribute.")
 
     def __getattribute__(cls, attr):
-        if attr in ('description', 'error_msg'):
+        if attr in ('description', 'error_msg') or attr.startswith('_'):
```
astrofrog: This makes sense anyway since the custom error is only meant to be shown for command arguments, which should be public.

astrofrog:
I think the other four failures are unrelated though, right? I think at least with this change, it will be more obvious the astropy-helpers tests are being run. That is, I completely agree we need the change in the template, but I think we should also include this as I think it's a good idea regardless of the pytest issue.

bsipocz:
OK, fair enough. I also agree that having failures instead of an error is an improvement 😅

PR #490, opened 2019-06-29
Correcty typo in error message

hamogu:
Without this fix, there is a space missing between "build" and "the"
[skip CI]


bsipocz:
Thanks @hamogu!

PR #493, opened 2019-07-23
Fix deprecation warning in theme

astrofrog:
Fixes https://github.com/astropy/astropy-helpers/issues/483


bsipocz:
Thanks @astrofrog! This wasn't the cause to the LTS doc build failure (but the missing pytest intersphinx), but nevertheless, having less deprecation warnings around is a good thing.

PR #495, opened 2019-08-27
Revert "Hack to pass -coverage to linker"

### LINKED ISSUE #432: OpenMP detection broken when building with -coverage
Detection of OpenMP support when building with the `-coverage` option (see #374) is broken again because the `CFLAGS` environment variable is propagated to the compiler invocation but not the linker invocation. The environment variable `CFLAGS=-coverage` normally tells the build process to add the `-coverage` option to both compiler AND linker commands for C binaries, but as shown below it is only passed to the compiler. As a result, the executable fails to run.

```
gcc -pthread -DNDEBUG -g -fwrapv -O3 -Wall -coverage -fPIC -c test_openmp.c -o objects/test_openmp.o -fopenmp
gcc -pthread objects/test_openmp.o -o test_openmp -fopenmp
objects/test_openmp.o: In function `_GLOBAL__sub_I_65535_0_test_openmp.c':
test_openmp.c:(.text.startup+0x38): undefined reference to `__gcov_init'
objects/test_openmp.o:(.data.rel+0x20): undefined reference to `__gcov_merge_add'
collect2: error: ld returned 1 exit status
```

lpsinger:
This reverts commit b4ffde3dc7bc30b93050e66c7288e79e3058433e.

It did not fix #432.

### DIFF
--- a/astropy_helpers/openmp_helpers.py
+++ b/astropy_helpers/openmp_helpers.py
@@ -174,11 +174,6 @@ def check_openmp_support(openmp_flags=None):
     compile_flags = openmp_flags.get('compiler_flags')
     link_flags = openmp_flags.get('linker_flags')
 
-    # Pass -coverage flag to linker.
-    # https://github.com/astropy/astropy-helpers/pull/374
-    if '-coverage' in compile_flags and '-coverage' not in link_flags:
-        link_flags.append('-coverage')
-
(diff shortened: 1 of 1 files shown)


pllim:
@bsipocz , I think we can go ahead and merge this one.  This removes the dead code that I had put in.

bsipocz:
Thanks @lpsinger!

Issue #492, opened 2019-07-12
INFO: v3.2.1 update resulted in deprecation warnings and undefined _err_help_msg


pllim:
I really like and appreciate the automated PRs to update helpers; Thank you very much!

However, when there is a large refactoring, it is almost unavoidable that an automated script would miss something. I am opening this issue in case others see similar issues after accepting the automated v3.2.1 update for `astropy-helpers` (e.g., spacetelescope/synphot_refactor#203).

You might see:

* Deprecation warnings from `astropy-helpers` when you run `setup.py install` of your package.
* Undefined variable called `_err_help_msg` when you run `flake8` on your `ah_bootstrap.py`.

Here is an example PR on how these can be fixed: spacetelescope/synphot_refactor#205

bsipocz:
The flake8 stuff is unfortunate, but I would suggest to ignore all external package from pep8 testing anyway.

pllim:
Well, `flake8` pointed it out to me, but undefined variable will cause exception if that logic block is run, so it is not just a stylistic preference. Granted that by the time that logic runs, you are already in trouble, but it would mask the real underlying problem.

bsipocz:
Would you mind PR it here, too?

pllim:
There is no problem here (I checked). It is the problem with the automated script used to patch packages, particularly the ones made with a very old template, but that ship has sailed.

bsipocz:
But that is weird. The script should copy `ah_bootstrap.py` directly from the helpers, as is in the release.

PR #496, opened 2019-09-11
Fixed undefined var in ah_bootstrap

pllim:
Part of #492 . It is defined too late in the file, causing `flake8` to complain about missing variable.


bsipocz:
Thanks @pllim! 

I'll deal with appveyor separately.

PR #498, opened 2019-11-13
Add a --parallel flag for build_docs

astrofrog:
This does several things:

* Adds a ``--parallel`` option for ``build_docs``
* Drops support for Python 3.5
* Adds a requirement on sphinx-astropy 1.2 - this is a very recent release, but I think probably ok since this is mostly for package maintainers/developers who can easily update?
* Requiring sphinx-astropy 1.2 also allows us to support only Sphinx 1.7 and above, so simplifies some of the logic here
* Removed leftover references to auto-installing sphinx-astropy
* Switch to Travis CI for Windows builds (I couldn't fix the AppVeyor one)


astrofrog:
Trying to debug the failing CI...

.travis.yml
```diff
@@ -23,41 +22,53 @@ env:
     - PIP_DEPENDENCIES="codecov"
     - EVENT_TYPE='push pull_request'
     - DEBUG=True
+    - CONDA_CHANNELS="astropy"
```
bsipocz: what do we install from conda that we need from the astropy channel?

bsipocz: e.g. could just add this to the one build that conda installs `sphinx-astropy` rather than being a global setting (probably doesn't matter much, but I expect things might be slightly be faster using just the default channels)

.travis.yml
```diff
@@ -23,41 +22,53 @@ env:
...
+    - os: windows
+      env: PYTHON_VERSION=3.7
+           CONDA_DEPENDENCIES="setuptools cython numpy pytest-cov"
+           PIP_DEPENDENCIES="codecov sphinx-astropy"
+
     # Test conda's clang
     - os: osx
       env:
-        - PYTHON_VERSION=3.5
+        - PYTHON_VERSION=3.7
```
bsipocz: we need python 3.8 builds, too

CHANGES.rst
```diff
@@ -4,7 +4,9 @@ astropy-helpers Changelog
 4.0 (unreleased)
 ------------------
 
-- No changes yet.
+- Changed minimum required Python version to 3.6. [#498]
```
bsipocz: would also mention the change in the minimum sphinx version

astropy_helpers/commands/build_sphinx.py
```diff
@@ -35,25 +31,17 @@
 def ensure_sphinx_astropy_installed():
     """
     Make sure that sphinx-astropy is available.
-
-    This returns the available version of sphinx-astropy as well as any
-    paths that should be added to sys.path for sphinx-astropy to be available.
     """
-    # We've split out the Sphinx part of astropy-helpers into sphinx-astropy
-    # but we want it to be auto-installed seamlessly for anyone using
```
bsipocz: probably worth mentioning in the changelog that there is no more auto installing

astropy_helpers/commands/build_sphinx.py
```diff
@@ -99,6 +92,7 @@ def initialize_options(self):
         self.open_docs_in_browser = False
         self.warnings_returncode = False
         self.traceback = False
+        self.parallel = None
```
bsipocz: do we maybe want some parallelism by default?

.travis.yml
astrofrog: Done

.travis.yml
astrofrog: I think we can do that in a separate PR since that's well beyond the scope of this one :)

CHANGES.rst
astrofrog: Done

astropy_helpers/commands/build_sphinx.py
astrofrog: This was actually done in https://github.com/astropy/astropy-helpers/pull/474 for 3.2 - I'm just removing leftover comments.

astropy_helpers/commands/build_sphinx.py
astrofrog: I'm not sure we want to do that since it's still somewhat experimental.  I'd rather leave it opt in if that's ok? (consistent with tests)

bsipocz:
Thanks @astrofrog!

PR #499, opened 2019-11-18
Adding python3.8 testing

bsipocz:
We need to be sure we're 3.8 compatible with the 4.0 release. RC can be done once this one is merged.


bsipocz:
Cancelled all but the 3.8 CI jobs, merge away once those pass.

bsipocz:
Apparently mpl is pulled in by sphinx-astropy, and it doesn't yet seems to be compatible, at least not on osx.

So I just go ahead and add the job for linux, but not for osx.

Issue #505, opened 2019-12-14
setup.py broken?


juliotux:
I don't know exactly why, but `astropy_helpers` present the same problem when installing via `pip` or `ah_bootstrap.py`. I tried manual install and got these results for release and git versions:

```
pip install astropy_helpers
Collecting astropy_helpers
  Downloading https://files.pythonhosted.org/packages/57/f1/9b11cbda11f320db80fba7e0c4a54d70d38eb3370bfec99ae22232df88c4/astropy-helpers-4.0.tar.gz (52kB)
     |████████████████████████████████| 61kB 1.1MB/s 
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  ERROR: Command errored out with exit status 1:
   command: /home/julio/anaconda3/bin/python /home/julio/anaconda3/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /tmp/tmp_dsc6wbh
       cwd: /tmp/pip-install-y0gan_5f/astropy-helpers
  Complete output (16 lines):
  Traceback (most recent call last):
    File "/home/julio/anaconda3/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 207, in <module>
      main()
    File "/home/julio/anaconda3/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 197, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/home/julio/anaconda3/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 54, in get_requires_for_build_wheel
      return hook(config_settings)
    File "/tmp/pip-build-env-z9riqrie/overlay/lib/python3.7/site-packages/setuptools/build_meta.py", line 146, in get_requires_for_build_wheel
      return self._get_build_requires(config_settings, requirements=['wheel'])
    File "/tmp/pip-build-env-z9riqrie/overlay/lib/python3.7/site-packages/setuptools/build_meta.py", line 127, in _get_build_requires
      self.run_setup()
    File "/tmp/pip-build-env-z9riqrie/overlay/lib/python3.7/site-packages/setuptools/build_meta.py", line 142, in run_setup
      exec(compile(code, __file__, 'exec'), locals())
    File "setup.py", line 17, in <module>
      from astropy_helpers.version_helpers import generate_version_py  # noqa
  ModuleNotFoundError: No module named 'astropy_helpers'
[...]

PR #506, opened 2019-12-14
Fix pip installs of astropy-helpers

astrofrog:
As described in https://github.com/astropy/astropy-helpers/issues/505, pip installs of astropy-helpers currently is broken, due to the addition of the ``pyproject.toml`` file which means we need to take an extra step to be able to import astropy-helpers in the ``setup.py`` file here.

This issue will affect:

* Anyone trying to pip install astropy-helpers (since there are no wheels). Arguably this is a corner case since in general astropy-helpers is not meant to be pip installed directly.

* Anyone using setuptools 42 and above who is installing a Python package from source that uses astropy-helpers 4.0. Again, this is a corner case since the only package using astropy-helpers 4.0 is the core package and that one will have wheels.

However I think for the second point we need to make sure we get this fix in and release astropy-helpers 4.0.1 before we upgrade astropy-helpers in other packages.

@bsipocz - this still needs a pull request but it looks like the current changelog is not up to date to reflect the 4.0 release - do you have some unpushed commits?

Fixes https://github.com/astropy/astropy-helpers/issues/505


pllim:
Looks like helpers is not going down without a fight!

bsipocz:
Hmm. pip installing astropy-helpers indeed fails for me locally, too. But no matter how hard I try I cannot make the astropy sdist install fail, so I suppose it's not critical to work this into the release right now. What do you say @astrofrog?

bsipocz:
btw, I've updated the changelog and setup files with the release

juliotux:
Now `pip install` runs without problem with this branch.

```
[julio@Dell7995 ~]$ conda activate
(base) [julio@Dell7995 ~]$ pip install -U https://github.com/astrofrog/astropy-helpers/archive/pip-install.zip
Collecting https://github.com/astrofrog/astropy-helpers/archive/pip-install.zip
  Downloading https://github.com/astrofrog/astropy-helpers/archive/pip-install.zip
     / 122kB 2.4MB/s
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Building wheels for collected packages: astropy-helpers
  Building wheel for astropy-helpers (PEP 517) ... done
  Created wheel for astropy-helpers: filename=astropy_helpers-4.1.dev0-cp37-none-any.whl size=56553 sha256=2ed6880770e83898d58c4e972ec8cb750105a05b966c68bb6bd47cac60712abd
  Stored in directory: /home/julio/.cache/pip/wheels/93/89/d0/6157655e60ccbd03a0d1654feb763bf13a2da6e541074382ec
Successfully built astropy-helpers
Installing collected packages: astropy-helpers
Successfully installed astropy-helpers-4.1.dev0
```

pllim:
A little ironic given we have just accepted [APE 17](https://github.com/astropy/astropy-APEs/blob/master/APE17.rst). 😅

astrofrog:
It's more a good illustration that more things are going to break in astropy-helpers due to changes in the Python packaging ecosystem if we don't move away from it! 😛

bsipocz:
Thanks @astrofrog

PR #511, opened 2020-05-20
Fix leaked temporary directory in check_openmp_support

lpsinger:

PR #512, opened 2020-05-20
TST: Fix failing CI

pllim:
Unrelated failures were seen at #511 , blocking it from proceeding.

The real problem: `sphinx-astropy` not found on `conda`.

Bonus: Remove or ignore warnings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,13 @@ env:
   matrix:
     - PYTHON_VERSION=3.6
     - PYTHON_VERSION=3.7 SETUPTOOLS_VERSION=dev DEBUG=True
-      CONDA_DEPENDENCIES="cython numpy pytest-cov sphinx-astropy'
+      CONDA_DEPENDENCIES="cython numpy pytest-cov sphinx-astropy>=1.2 matplotlib>=3.1"
+      CONDA_CHANNELS="astropy"
       EVENT_TYPE='push pull_request cron'
 
   global:
-    - CONDA_DEPENDENCIES="setuptools sphinx-astropy cython numpy pytest-cov"
+    - CONDA_DEPENDENCIES="setuptools sphinx-astropy>=1.2 cython numpy pytest-cov"
+    - CONDA_CHANNELS="astropy"
     - PIP_DEPENDENCIES="codecov"
     - EVENT_TYPE='push pull_request'
     - DEBUG=True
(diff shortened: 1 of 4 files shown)


pllim:
I actually don't know if after moving sphinx-astropy from conda to pip install for all jobs, now there are actually duplicate jobs or not... Someone familiar with the test matrix should look at this.

pllim:
@astrofrog , I reverted the change to force grab sphinx-astropy from pip but the first job fails because conda insists on downgrading sphinx-astropy to 1.1.1 even when you need 1.2. How do you want me to fix it without forcing pip install? Please advise. Thanks.


```
sphinx-astropy-1.1.1       |     pyh734ff71_0          18 KB  astropy
```

astrofrog:
I wonder if conda is somehow too old. Can you try setting ``MINICONDA_VERSION`` to ``latest``?

astrofrog:
I'm also going to ping @mwcraig as our resident conda expert 😁

pllim:
`latests` made things worse, as now the Linux job fails too. Previously it only fails on OSX, that first job. 🤷

mwcraig:
I've looked at this a couple of times, no ideas yet. Would  it work to specify a minimum version for `sphinx-astropy`?

FWIW, I think `ci-helpers` needs a big refactor, but that is outside the scope of this issue.

mwcraig:
Looks like it is related to this https://github.com/sphinx-doc/sphinx/issues/7659; the error is no longer that the wrong version of sphinx-astropy is being used, it is that an older matplotlib is being used. I'm not sure *why* that is on the mac, but can you try adding `matplotlib >3` to the failing OSX jobs and see what happens?

pllim:
Eh... where did the CI go?

mwcraig:
It looks like the matplotlib version was added to the wrong build, but tbh just removing that build seems fine to me.

pllim:
Who uses Macs anyways? 😉

pllim:
What....? I just can't make conda happy.

pllim:
Just for the record, my very first attempt to just pip install `sphinx-astropy` outside of `conda` got all green.

pllim:
I think I am giving up unless there are more ideas I can try...

mwcraig:
Let's just go with the pip install...looks like the latest issue is related to coverage, where the the restriction `coverage <5` is not being respected by conda (well, really, the issue is that the pinning mechanism used in ci-helpers no longer works in conda)

astrofrog:
Before we go with the pip install route can you try and set MAMBA=True in the Travis config as this will then try and use mamba instead of conda which can resolve dependencies better and give more explicit errors?

astrofrog:
(This is a recent ci-helpers addition)

pllim:
I think #515 superseded this.

PR #515, opened 2020-09-13
Fix CI failures

astrofrog:
Use mamba to avoid conda issues that are causing CI failures (https://github.com/astropy/astropy-helpers/pull/512) - mamba is generally better with any kind of conflicts etc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ env:
       EVENT_TYPE='push pull_request cron'
 
   global:
+    - MAMBA=True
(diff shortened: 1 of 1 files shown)


astrofrog:
I'm pretty sure gcc is no longer supported in conda on MacOS X hence why the package is now missing (had to remove the MacOS X gcc build, as we did in extension-helpers)

pllim:
Since this repo is pretty much legacy, whatever works is fine. I can't merge though but :+1 FWIW.

